Issue #4, opened 2020-05-03
getOrCreateIndex method


curquiza:
Issue fully detailed [here](https://github.com/meilisearch/integration-guides/issues/2).

ppamorim:
@curquiza Could you please review the PR?

PR #7, opened 2020-05-06
Added getOrCreateIndex function

ppamorim:
Implements the #4. Please do not merge yet due missing documentation.

### DIFF
--- a/Sources/MeiliSearch/Indexes.swift
+++ b/Sources/MeiliSearch/Indexes.swift
@@ -84,29 +84,29 @@ struct Indexes {
     enum CreateError: Swift.Error {
         case indexAlreadyExists
 
-      static func decode(_ error: MSError) -> Swift.Error {
+        static func decode(_ error: MSError) -> Swift.Error {
 
-        let underlyingError: NSError = error.underlying as NSError
+            let underlyingError: NSError = error.underlying as NSError
 
-        if let data = error.data {
+            if let data = error.data {
 
-          let msErrorResponse: MSErrorResponse?
-          do {
-              let decoder: JSONDecoder = JSONDecoder()
-              msErrorResponse = try decoder.decode(MSErrorResponse.self, from: data)
-          } catch {
-              msErrorResponse = nil
-          }
+                let msErrorResponse: MSErrorResponse?
+                do {
+                    let decoder: JSONDecoder = JSONDecoder()
+                    msErrorResponse = try decoder.decode(MSErrorResponse.self, from: data)
+                } catch {
+                    msErrorResponse = nil
+                }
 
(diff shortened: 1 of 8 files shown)


Tests/MeiliSearchTests/MockURLSession.swift
```diff
@@ -57,3 +72,13 @@ class MockURLSessionDataTask: URLSessionDataTaskProtocol {
         resumeWasCalled = true
     }
 }
+
+private let ERROR = 0
+private let SUCCESS = 1
```
curquiza, line 77: Is it not the contrary?
In general, we return 0 when succeeded, and 1 when failed.

ppamorim: You are correct, I was thinking about `Bool(1) == true` and `Bool(0) == false`.

curquiza: Haha, I understand the confusion 😉

ppamorim:
@qdequele Yes I am, are you asking about the number of tabs? If yes I can see where you are coming from, there is a file with 1 tab instead of one, I will fix it. The Swiftlint linter does not detect this I think.

Tests/MeiliSearchTests/MockURLSession.swift
curquiza: So, can you change that @ppamorim? 
```
ERROR = 1
SUCCESS = 0
```
or
```
TRUE = 1
FALSE = 0
```

ppamorim: @curquiza I ditch this and implemented a profer solution using the enum `ResponseStatus` with `success and `error`.

curquiza: I like your solution!

Sources/MeiliSearch/Indexes.swift
```diff
@@ -58,6 +58,57 @@ struct Indexes {
...
+                    self.get(UID, completion)
+                default:
+                    completion(.failure(error))
+                }
+
+            }
+
+        }
+
+    }
```
qdequele, line 82: @ppamorim No, I tell that because I see sometimes an empty line behind a scope close bracket. Like on these lines.

ppamorim: From where this rule is coming from? `swiftlint` does not complain about it. If you want you can open an issue that we can fix it since this is all over the project.

Please check this out: https://google.github.io/swift/#braces

curquiza: @qdequele :)

qdequele: Thanks @ppamorim. It's not a problem. I'm sorry I slowed things down.

ppamorim:
Thank you all!

PR #13, opened 2020-05-29
Prepare for MeilisSearch v0.11.0 : Use new error handler to check if index already exists

curquiza:
It must be merged once the MeiliSearch `v0.11.0` is released. This version includes `errorCode`, `errorType` and `errorLink` in answer body.

Warning:
The answer body still includes a `message` attribute, but the error message for "This index already exists" error will change. So this PR has to be merged as quickly as possible once the v11 is out because this package will not work with the v11 without this fix.
So it's normal that the tests fail right now.


ppamorim:
@curquiza I will have a look on this.

PR #21, opened 2020-07-06
Rename addOrReplaceDocument and addOrUpdateDocument into addDocuments and updateDocuments.

curquiza:
Rename `addOrReplaceDocument` and `addOrUpdateDocument` into `addDocuments` and `updateDocuments`.

With the team, we have chosen to call the documents methods (in the SDKs) `addDocuments` for `addOrReplaceDocuments` and `updateDocuments` for `AddOrUpdateDocuments`.
Why? For simplicity. It’s clearer for the user without reading/understanding the MeiliSearch docs. You want to add: you use `addDocuments`. You want to update: you use `updateDocuments`. An API has to be accurate in its routes naming. An SDK has to be easy to use since it's a wrapper to help user play and code with MeiliSearch 🚀
That’s why we decided to keep these shorter names 🙂

We already use these names in all our other SDKs. What do you think about the Swift one @ppamorim? 🙂

In any case, this PR is needed since it fixes a typo (`addOrReplaceDocument` has currenlty no `s` at the end).

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,8 +124,8 @@ The current functions implemented are:
 
 | Swift function  | Similar to API call |
 | ------------- | ------------- |
-| `addOrReplaceDocument(UID:document:primaryKey:_)`  | `POST - /indexes/:index_uid/documents`  |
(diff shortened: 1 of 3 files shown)


ppamorim:
Thank you for the PR! Merging it now.

PR #44, opened 2020-08-18
Indexes tests

bidoubiwa:
Tests added to test the index routes against meilisearch

## Failing tests
- updateIndex
Fails because of #45 #47


curquiza:
@bidoubiwa, why is this PR closed?

PR #51, opened 2020-10-02
[#48] Remove acceptNewFields routes

### LINKED ISSUE #48: Remove acceptNewFields
According to this issue https://github.com/meilisearch/integration-guides/issues/39

Ethenyl:
# Summary

Fixes #48.

Remove `acceptNewFields` routes and add default values for `searchableAttributes` and `displayedAttributes` in `Settings`.

# Miscellaneous

Added UT for testing default values and removed `distinctAttributes` from JSON test payload.

### DIFF
--- a/Sources/MeiliSearch/Model/Setting.swift
+++ b/Sources/MeiliSearch/Model/Setting.swift
@@ -3,53 +3,42 @@ import Foundation
 /**
  Each instance of MeiliSearch has three keys: a master, a private, and a public.
  Each key has a given set of permissions on the API routes.
-*/
+ */
 public struct Setting: Codable, Equatable {
 
     // MARK: Properties
 
-    ///List of ranking rules for a given `Index`.
+    /// List of ranking rules for a given `Index`.
     public let rankingRules: [String]
 
-    ///List of searchable attributes for a given `Index`.
+    /// List of searchable attributes for a given `Index`.
     public let searchableAttributes: [String]
 
-    ///List of displayed attributes for a given `Index`.
+    /// List of displayed attributes for a given `Index`.
     public let displayedAttributes: [String]
 
-    ///List of stop-words for a given `Index`.
+    /// List of stop-words for a given `Index`.
     public let stopWords: [String]
 
-    ///List of synonyms and its values for a given `Index`.
+    /// List of synonyms and its values for a given `Index`.
     public let synonyms: [String: [String]]
 
-    ///Return if a given `Index` allows new fields.
-    public let acceptNewFields: Bool
-
-    public init(
-      rankingRules: [String],
-      searchableAttributes: [String],
-      displayedAttributes: [String],
-      stopWords: [String],
-      synonyms: [String: [String]],
-      acceptNewFields: Bool) {
-      self.rankingRules = rankingRules
-      self.searchableAttributes = searchableAttributes
-      self.displayedAttributes = displayedAttributes
-      self.stopWords = stopWords
-      self.synonyms = synonyms
-      self.acceptNewFields = acceptNewFields
+    public init(rankingRules: [String], searchableAttributes: [String], displayedAttributes: [String], stopWords: [String], synonyms: [String: [String]]) {
+        self.rankingRules = rankingRules
+        self.searchableAttributes = searchableAttributes
+        self.displayedAttributes = displayedAttributes
+        self.stopWords = stopWords
+        self.synonyms = synonyms
     }
 
-    ///Tries to decode the JSON object to Setting object.
+    /// Tries to decode the JSON object to Setting object.
     public init(from decoder: Decoder) throws {
         let values = try? decoder.container(keyedBy: CodingKeys.self)
+
         rankingRules = (try? values?.decodeIfPresent([String].self, forKey: .rankingRules)) ?? []
-        searchableAttributes = (try? values?.decodeIfPresent([String].self, forKey: .searchableAttributes)) ?? []
-        displayedAttributes = (try? values?.decodeIfPresent([String].self, forKey: .displayedAttributes)) ?? []
+        searchableAttributes = (try? values?.decodeIfPresent([String].self, forKey: .searchableAttributes)) ?? ["*"]
+        displayedAttributes = (try? values?.decodeIfPresent([String].self, forKey: .displayedAttributes)) ?? ["*"]
         stopWords = (try? values?.decodeIfPresent([String].self, forKey: .stopWords)) ?? []
         synonyms = (try? values?.decodeIfPresent([String: [String]].self, forKey: .synonyms)) ?? [:]
-        acceptNewFields = (try? values?.decodeIfPresent(Bool.self, forKey: .acceptNewFields)) ?? false
     }
-
 }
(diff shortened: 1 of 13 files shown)


Sources/MeiliSearch/Model/Setting.swift
```diff
@@ -3,53 +3,42 @@ import Foundation
...
-      stopWords: [String],
-      synonyms: [String: [String]],
-      acceptNewFields: Bool) {
-      self.rankingRules = rankingRules
-      self.searchableAttributes = searchableAttributes
-      self.displayedAttributes = displayedAttributes
-      self.stopWords = stopWords
-      self.synonyms = synonyms
-      self.acceptNewFields = acceptNewFields
+    public init(rankingRules: [String], searchableAttributes: [String], displayedAttributes: [String], stopWords: [String], synonyms: [String: [String]]) {
```
ppamorim, line 26: Is this init needed since `Setting` is a struct?

Ethenyl: Nope it can be completely removed, I will do it right away.

Ethenyl: Removed in `395e888`.

ppamorim (approved):

ppamorim:
@curquiza Could you please review and merge if approved?

Tests/MeiliSearchUnitTests/SettingsTests.swift
```diff
@@ -17,7 +17,6 @@ class SettingsTests: XCTestCase {
             "exactness",
             "desc(release_date)"
         ],
-        "distinctAttribute": null,
```
bidoubiwa, line 20: Why is this removed ?

Tests/MeiliSearchUnitTests/SettingsTests.swift
```diff
@@ -42,11 +40,21 @@ class SettingsTests: XCTestCase {
...
+        XCTAssertTrue(stubSetting.stopWords.isEmpty)
+        XCTAssertTrue(stubSetting.synonyms.isEmpty)
+    }
+
     func testGetSetting() {
 
         //Prepare the mock server
 
-        let stubSetting: Setting = buildStubSetting()
+        let stubSetting = buildStubSetting(from: json)
```
bidoubiwa, line 57: Why is the typing removed ?

Tests/MeiliSearchUnitTests/SettingsTests.swift
```diff
@@ -42,11 +40,21 @@ class SettingsTests: XCTestCase {
 
     // MARK: Settings
 
+    func testShouldInstantiateFromEmptyData() {
```
bidoubiwa, line 43: Where is this function called ?

Sources/MeiliSearch/Model/Setting.swift
```diff
@@ -3,53 +3,34 @@ import Foundation
...
-      self.stopWords = stopWords
-      self.synonyms = synonyms
-      self.acceptNewFields = acceptNewFields
-    }
-
-    ///Tries to decode the JSON object to Setting object.
+    /// Tries to decode the JSON object to Setting object.
     public init(from decoder: Decoder) throws {
         let values = try? decoder.container(keyedBy: CodingKeys.self)
+
```
bidoubiwa, line 29: Seems like distinctAttribute is missing here, should a PR be opened to fix this @ppamorim ?

bidoubiwa (commented):
Thanks a lot for this contribution! It's a really good work :)

Tests/MeiliSearchUnitTests/SettingsTests.swift
ppamorim: @bidoubiwa Yes, we would like to keep the types explicit.

Tests/MeiliSearchUnitTests/SettingsTests.swift
Ethenyl: I removed since there's no usage of it whatsoever, if you think it is still relevant, I'll revert this change.

Tests/MeiliSearchUnitTests/SettingsTests.swift
Ethenyl: No issue, what is the reason behind this requirement?

I tend to remove any duplication wherever I can, less code = less headaches but I can revert this if needed.

Tests/MeiliSearchUnitTests/SettingsTests.swift
Ethenyl: It is a test, test are automatically detected and added by Xcode, no need to call them manually.

But I see your point, for Linux I need to add the test manually at the end of the file, I'm not used to that, I never perform test for Linux.

Tests/MeiliSearchUnitTests/SettingsTests.swift
ppamorim: This is needed to enforce that types are the same as expected (question yourself: what's the return type of the function `buildStubSetting` without the need of going to the function?), also it helps Xcode to index the code faster, we suffer for constant problems where Xcode is not capable of indexing it faster than a human. :upside_down_face:

Tests/MeiliSearchUnitTests/SettingsTests.swift
bidoubiwa: I did not know that. I'm used to adding them at the end of the file mostly also to have a quick way of knowing which tests are done in each file

Tests/MeiliSearchUnitTests/SettingsTests.swift
bidoubiwa: You are right, it is missing.

Tests/MeiliSearchUnitTests/SettingsTests.swift
Ethenyl: Ok, I never encountered those kind of issues before, even on far bigger projects (which mixed ObjC and Swift) but I will follow your coding guidelines.

As for the naming of the function, it should be called getStubSetting to say that it actually returns this 😉

Tests/MeiliSearchUnitTests/SettingsTests.swift
Ethenyl: Well, I forgot to add, functions prefixed with `test` will be counted as tests 😄

Tests/MeiliSearchUnitTests/SettingsTests.swift
Ethenyl: Fixed in `0eeac0c`.

Ethenyl:
Just FYI, I had to reword my last commit since the build failed to restart the build.

And without changing anything, the build completed successfully this time, the failing test was `testDeleteDocument` in `DocumentsTests.swift`.

ppamorim:
@Ethenyl Thank you very much, we really appreciate your effort in this PR.

Regarding the failing `testDeleteDocument` test, don't worry, we are sorting these tests in the PR #54.

ppamorim (approved):

bidoubiwa (approved):
Thanks a lot

curquiza (approved):

Issue #37, opened 2020-08-11
Should not ask user to jsonEncode when using the SDK


bidoubiwa:
## Problem

When using the SDK, before each call an object must be encoded in JSON. But the returned object are already decoded into the expected type. This lacks in coherence

### Exemple

In the following example I had to JSONEncode my `movie` into the `Data` type before adding it to the document parameter of `addDocuments`.
When In the second request I use the method `getOneDocument`, I recieve my movie back not as a `Movie` type and not as a `Data` type. 

It should either be Data to Data, or Swift type to Swift type.

```swift
    func testAddAndGetOneDocuments() {
        let movie: Movie = Movie(id: 10, title: "test", comment: "test movie")
        let documents: Data = try! JSONEncoder().encode([movie])

        let expectation = XCTestExpectation(description: "Add or replace Movies document")

        self.client.addDocuments(
            UID: uid,
            documents: documents,
            primaryKey: nil
        ) { result in
            switch result {
            case .success(let update):
                XCTAssertEqual(Update(updateId: 0), update)
                expectation.fulfill()
            case .failure(let error):
                XCTFail(error.localizedDescription)
            }
        }
        self.wait(for: [expectation], timeout: 1.0)
        sleep(1)
        let getExpectation = XCTestExpectation(description: "Add or replace Movies document")
        self.client.getDocument(
            UID: uid,
            identifier: "10"
        ) { (result: Result<Movie, Swift.Error>) in

            switch result {
            case .success(let returnedMovie):
                print("HELLO")
                print(returnedMovie)
                XCTAssertEqual(movie, returnedMovie)
                getExpectation.fulfill()
            case .failure(let error):
                XCTFail(error.localizedDescription)
            }
        }
        self.wait(for: [getExpectation], timeout: 3.0)
    }
```

## Solution

I prefer that the user sends its data using its specific type (in this case `Movie`). Swift will most likely be used for front end matter so while I understand that it makes it more complex to add a "big json" to meilisearch, i'm not sure any user will use swift to add documents to meilisearch (maybe i'm wrong). And in the rare cases it does, the data will come from whatever `form` the user completed and thus the data will be in a given format (i.e: Movie) and not in a jsonString.

In this case, the function prototype will look like this:

```
public func addDocuments<T>(
        UID: String,
        documents: T,
        primaryKey: String?,
        _ completion: @escaping (Result<Update, Swift.Error>) -> Void) {
        self.documents.addOrReplace(
            UID,
[...]

PR #53, opened 2020-10-04
Add support to Encodable types for addDocuments function

ppamorim:
## Summary

- Function `addDocuments<T>` added. Now it's possible to send objects that are `Codable/Encodable` and the SDK will automatically encode the values to JSON. A custom encoder can be set by adding `encoder` in the call.

## Additions

- Fixed bug in the `ClientTests.swift` file

## Local tests

 - [X] Unit tests passed
 - [X] Integration tests passed

## Problems

Users are forced to wrap the object to `[]` when using `addDocuments` function with objects that are `Codable/Encodable`. I tried to create a wrapper function for this but Swift doesn't like the direct conversion from `T` to `[T]`.

### DIFF
--- a/Sources/MeiliSearch/Client.swift
+++ b/Sources/MeiliSearch/Client.swift
@@ -143,7 +143,36 @@ public struct MeiliSearch {
      For a partial update of the document see `updateDocument`.
 
      - parameter UID:        The unique identifier for the Document's index to be found.
-     - parameter documents:  The documents data (JSON) to be processed.
+     - parameter documents:  The documents to be processed.
+     - parameter completion: The completion closure used to notify when the server
+     completes the update request, it returns a `Result` object that contains `Update`
+     value. If the request was sucessful or `Error` if a failure occured.
+     */
+    public func addDocuments<T>(
+        UID: String,
+        documents: [T],
+        encoder: JSONEncoder? = nil,
+        primaryKey: String?,
+        _ completion: @escaping (Result<Update, Swift.Error>) -> Void) where T: Encodable {
+        self.documents.add(
+            UID,
+            documents,
+            encoder,
+            primaryKey,
+            completion)
+    }
+
+    /**
+     Add a list of documents as data or replace them if they already exist.
+
+     If you send an already existing document (same id) the whole existing document will
+     be overwritten by the new document. Fields previously in the document not present in
+     the new document are removed.
+
+     For a partial update of the document see `updateDocument`.
+
+     - parameter UID:        The unique identifier for the Document's index to be found.
+     - parameter documents:  The  data to be processed.
      - parameter completion: The completion closure used to notify when the server
      completes the update request, it returns a `Result` object that contains `Update`
      value. If the request was sucessful or `Error` if a failure occured.
--- a/Sources/MeiliSearch/Constants.swift
+++ b/Sources/MeiliSearch/Constants.swift
@@ -7,4 +7,11 @@ struct Constants {
         decoder.dateDecodingStrategy = .formatted(Formatter.iso8601)
         return decoder
     }()
+
+    static let customJSONEecoder: JSONEncoder = {
+        let encoder = JSONEncoder()
+        encoder.dateEncodingStrategy = .formatted(Formatter.iso8601)
+        return encoder
+    }()
+
 }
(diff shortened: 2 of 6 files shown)


curquiza:
Is this PR ready for review? Because it's still a draft 🙂

ppamorim:
@curquiza This is waiting for the PR #54.

curquiza:
bors try

curquiza:
Tell me when this draft is ready for review 🙂

ppamorim:
bors try

ppamorim:
bors try

bidoubiwa (commented):
Could you make a test where we add documents without jsonEncode them? Something like this

```swift
func testAddDataAndGetDocuments() {
      let expectation = XCTestExpectation(description: "Add or replace Movies document")

      self.client.addDocuments(
          UID: self.uid,
          documents: movies, //We use movies directly
          primaryKey: nil
      ) { result in
       ...
```

curquiza (commented):
@ppamorim Two questions:

- If I understand, we add another `addDocuments` method additionally to the previously existing one. Does this mean you can use the `addDocuments` method with or without the encodable support? Maybe I don't understand well, but what the point for the user to use a function that does not handle encodable types? I mean, should we let both of the method `addDocuments`?

- Should we add this functionality to the `updateDocuments` method? Not mandatory in this PR, but at least an issue should be created. The #37 is maybe not really accurate. In general, are there any other functions concerned by this user-experience improvement? @bidoubiwa

ppamorim:
> @ppamorim Two questions:
> 
> * If I understand, we add another `addDocuments` method additionally to the previously existing one. Does this mean you can use the `addDocuments` method with or without the encodable support? Maybe I don't understand well, but what the point for the user to use a function that does not handle encodable types? I mean, should we let both of the method `addDocuments`?
> * Should we add this functionality to the `updateDocuments` method? Not mandatory in this PR, but at least an issue should be created. The #37 is maybe not really accurate. In general, are there any other functions concerned by this user-experience improvement? @bidoubiwa

Answers in order:

[0]:

 - We added the support of codable types in the `addDocuments<T: Codable>(...)`, this function requires that the user passes an array of `T` that implements `Codable`. If you want to directly send the data buffer to the server, you can use the function `addDocuments(UID:documents:...)` where `documents` is the type data. This is especially useful if you are adding data without the need to transforming the data to an array of `Codable` and them passing the data to the JSON encoder, this is an overhead if you already have the JSON data ready for MeiliSearch server.

 Ex:

This code gets outdated:

```swift
let movie: Movie = Movie(id: 10, title: "test", comment: "test movie")
let documents: Data = try! JSONEncoder().encode([movie])
self.client.addDocuments(
            UID: uid,
            documents: documents,
            primaryKey: nil) { ... }
```

Now the user is capable of simply sending the values directly:

```swift
self.client.addDocuments(
            UID: uid,
            documents: [movie],
            primaryKey: nil) { ... }
```

The library will do all the `JSONEncoder` part automatically.

[1]:

 - Yes I agree that this should be implemented too, I will open a PR for this later, is that ok?

ppamorim:
@curquiza Missed to inform that if the user contains requires a custom date format, it can be passed as an argument by using the parameter `JSONEncoder`.

curquiza:
Yes I understand why what you implemented is useful, I don't understand why we keep the outdated function in this case:

```swift
let movie: Movie = Movie(id: 10, title: "test", comment: "test movie")
let documents: Data = try! JSONEncoder().encode([movie])
self.client.addDocuments(
            UID: uid,
            documents: documents,
            primaryKey: nil) { ... }
```

This version of `addDocuments` is outdated as you said, can we remove it?

ppamorim:
This is not outdated because they use may want to pass the data directly without the need of writing a `Codable` object for it. It will be used in the *** app.

curquiza:
Let's keep both then.

But, as @bidoubiwa said, could you add a test for that: https://github.com/meilisearch/meilisearch-swift/pull/53#pullrequestreview-506408681

> Yes I agree that this should be implemented too, I will open a PR for this later, is that ok?

Yes, opening an issue is better!

curquiza (approved):
LGTM 👌

curquiza:
bors merge

PR #32, opened 2020-08-06
add document and Client tests tested with a running meilisearch in CI

bidoubiwa:
## In this PR

- Indexes tests
- Documents tests
- Client tests

## Additions

- Tests on client and documents are now connected to MeiliSearch
- Major bug fix that made it hard to make some requests
- Added a CI to run tests againt MeiliSearch

PR #30, opened 2020-08-04
Change search method from get to post

bidoubiwa:
Not correctly tested yet but this will come in a later PR.

fixes: #26

PR #54, opened 2020-10-04
Create integrations tests for all MeiliSearch http routes 

ppamorim:
# Summary

- [x] Client tests  #32
- [x] Indexes tests #32 
- [x] Documents tests #32
- [x] add Settings integrations tests
   - [x] attributes for faceting tests
   - [x] displayed attributes tests
   - [x] distinct attributes tests
   - [x] ranking rules tests
   - [x] searchable attributes tests
   - [x] settings route tests
   - [x] stop words tests
   - [x] synonyms tests
- [x] update route tests
- [X] search route tests 
   - [X] limit tests
   - [X] offset tests
   - [X] attributes to crop tests
   - [X] cropLength tests
   - [X] matches tests
   - [X] attributesToHighlight tests
   - [X] attributesToRetrieve tests
   - [X] filters tests
   - [X] facetsDistribution tests
   - [X] facetFilters tests

## Reasons why the tests are failing

~~On the test `testGetDistinctAttribute` and `testUpdateDistinctAttribute`:~~

 ~~- The attribute that returns from the server contains a double "", causing the error: `XCTAssertEqual failed: ("product_id") is not equal to (""product_id"")`~~

 ~~On the test `testGetSettings` and `testUpdateSettings`:~~

 ~~- Probably due to the `acceptNewFields`, this has been removed.~~


ppamorim:
@curquiza Please review the tests, I am not sure what to do with the failed tests. Please bear that this PR is still a draft.

ppamorim:
@curquiza Waiting for the merge of #51 This will sort the settings test failure.

curquiza:
Hello @ppamorim! Thanks for this awesome PR!
The #51 is removed!

You can now [rebase your branch](https://gist.github.com/curquiza/5f7ce615f85331f083cd467fc4e19398) 🙂

ppamorim:
@curquiza Just did this.

curquiza:
awesome @ppamorim. If your PR is ready for review, can you remove the draft status of this PR? 🙂
We will review it asap!

ppamorim:
Waiting for PR #30 be rebased and merged.

ppamorim:
@curquiza It's not clear why the test `testSearchLimitEmptySearch` is failing in the CI. It's working locally and in the Dockerfile.

bidoubiwa:
@ppamorim is it possible to print the error message returned by MeiliSearch or by Swift ? It is hard to understand why a test fails without a error handler.

```
/Users/runner/work/meilisearch-swift/meilisearch-swift/Tests/MeiliSearchIntegrationTests/SearchTests.swift:222: error: -[MeiliSearchIntegrationTests.SearchTests testSearchLimitEmptySearch] : XCTAssertTrue failed
```

ppamorim:
It's possible by replacing `XCTAssertTrue(documents.hits.count == 0)` with `XCTAssertEqual(0, documents.hits)`. In the test `documents.hits` is returning 5. When running the very same test locally it returns 0 and the test passes.

ppamorim:
@bidoubiwa It doesn't make sense. I am running Meili on macOS.

Screenshot 2020-10-06 at 12 40 09

ppamorim:
Issue has been found, it was my local version of Meili, it was outdated. The behaviour of empty searches has been changed on version `v0.15`.

curquiza:
Oh! So sorry, it was my next task on my TODO to update these tests (of an empty query)! Sorry that it wastes your time!

ppamorim:
@curquiza No problem. All tests are passing, just waiting for the review.

ppamorim:
@curquiza Sorry I had to rebase it since Github informed that it was out of sync with the master.

curquiza:
> @curquiza Sorry I had to rebase it since Github informed that it was out of sync with the master.

Yes, GitHub will inform you, but Bors will do it at your place before merging 🙂 (except if you have merge conflicts. In this case you would have to rebase yourself)
So no worries about this GitHub warning.

ppamorim:
@curquiza @bidoubiwa Just waiting for the review now. :)

Sources/MeiliSearch/Model/SearchParameters.swift
```diff
@@ -82,15 +82,57 @@ public struct SearchParameters: Codable, Equatable {
...
-        value += "]"
-        return value
+}
+
+extension SearchParameters {
+
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(query, forKey: .query)
+        if limit != 20 {
```
curquiza, line 106: Why should we need to do a condition with the default value?

Tests/MeiliSearchIntegrationTests/XCTestManifests.swift
```diff
@@ -13,7 +13,9 @@ import XCTest
 #if !os(macOS)
 public func allTests() -> [XCTestCaseEntry] {
     [
+        testCase(IndexesTests.allTests),
         testCase(DocumentsTests.allTests),
+        testCase(SearchTests.allTests)
```
curquiza, line 18: Why the SettingsTests aren't there?

curquiza (changes requested):
Hello @ppamorim!
Huge work!

This is the review part I 😁 I will review the test files tomorrow!

One change asked 😇
- [ ] If I'm not wrong, there is no test about the `facetsDistribution` of `SearchResult`. I mean, doing a test with `facetsDistribution` as a search parameter and check that the `facetsDistribution` in the serach **result* are accessible. See the issue requirement (#58).

Tests/MeiliSearchIntegrationTests/XCTestManifests.swift
ppamorim: Missed the configuration.

Sources/MeiliSearch/Model/SearchParameters.swift
ppamorim: This is a filter to default parameters, so far I know these values are fixed in the server and if they are not informed in the HTTP request the server will apply the default ones, am I correct?

ppamorim:
> Hello @ppamorim!
> Huge work!
> 
> This is the review part I 😁 I will review the test files tomorrow!
> 
> One change asked 😇
> 
> * [ ]  If I'm not wrong, there is no test about the `facetsDistribution` of `SearchResult`. I mean, doing a test with `facetsDistribution` as a search parameter and check that the `facetsDistribution` in the serach *_result_ are accessible. See the issue requirement (#58).

There is a test related to this, it's named `testSearchFacetsDistribution`. It checks for the genres of books.

Sources/MeiliSearch/Model/SearchParameters.swift
ppamorim: The API is based in this documentation: https://docs.meilisearch.com/references/search.html#search-in-an-index-with-post-route

Note how the first example only contains the query item as mandatory, all other options are optional. This script remove the need of informing the default values.

curquiza:
> There is a test related to this, it's named testSearchFacetsDistribution. It checks for the genres of books and compares the dictionary that returns from Meilisearch.

My bad! I've seen it!

Sources/MeiliSearch/Model/SearchParameters.swift
curquiza: But what is the point of doing this?

> they are not informed in the HTTP request the server will apply the default ones,

Yes, exactly, so no need to check the default value. Just send to MeiliSearch what the users fill, we don't care if they fill default values. MeiliSearch will manage that.

This big `if` conditions are really complex for a job that MeiliSearch already does.

Or maybe I missed something!

Sources/MeiliSearch/Model/Stat.swift
```diff
@@ -16,6 +16,12 @@ public struct AllStats: Codable, Equatable {
     /// Dictionary of all Indexes containing the stat for each Index.  
     public let indexes: [String: Stat]
 
+    enum CodingKeys: String, CodingKey {
```
curquiza, line 19: If I'm not wrong a `CodingKeys` is for re-mapping the keys, only if you want to use a different key name in your code and in the JSON key.
That's not the case here because you are not redefined anything, for example: `case databaseSize = "DBSize"`.

ppamorim: That's the result of auto programming code. All  unnecessary`CodingKeys` were removed.

Tests/MeiliSearchIntegrationTests/UpdatesTests.swift
```diff
@@ -0,0 +1,148 @@
...
+
+                self.client.getUpdate(UID: self.uid, update) { result in
+
+                    switch result {
+                    case .success(let update):
+
+                        if update.status == "enqueued" {
+                          getUpdate()
+                          return
+                        }
```
curquiza, line 81: I'm not sure to understand what's the goal here?
Is it a recursive call? I would rather not have recursivity in the test part because it could do a stack overflow if MeiliSearch does not succeed to process the task.
Only checking you got a status, even if it's "enqueued" it's ok because we are only testing the `getUpdate` function, and not the `addDocuments` one 😉

curquiza (commented):
**Review part II**

You did a so HUGE work on the tests!!!! We all know here how it's a painful job to write tests and you created a lot! Thanks for that 🙂

Some remarks:
- still the questioning here (https://github.com/meilisearch/meilisearch-swift/pull/54#discussion_r500953007)
- I remark, for the **Settings tests** , for the `get` methods, you use the `update` method. For me, it's a little bit redundant with the `update` test right after: you can only check the default value of the settings.
But, as it's not really mandatory and just a detail, **I'm going to open a PR to your branch to suggest changes** so that it avoids you another painful work, just let me the day 😉

Sources/MeiliSearch/Model/SearchResult.swift
```diff
@@ -16,10 +17,71 @@ public struct SearchResult<T>: Codable, Equatable where T: Codable, T: Equatable
...
     public let limit: Int
 
+    /// Total number of matches,
+    public let nbHits: Int
+
+    /// Whether `nbHits` is exhaustive.
+    public let exhaustiveNbHits: Bool?
+
+    /// Distribution of the given facets.
+    public let facetsDistribution: [String: [String: Int]]?
```
curquiza, line 27: Can't you use a custom type `Facet` that is also Codable to avoid the Behemot? 🙂

```swift
public let facetsDistribution: [String: [Facet]]?
```

ppamorim: I don't think it's possible to map this, especially knowing that the structure is like this:

```swift
"genres": [
    "Classic Regency nove": 1,
    "High fantasy‎": 1,
    "Fantasy": 2,
    "Novel": 2,
    "Bildungsroman": 1
]
```

By creating the `Facet` type, we need to know beforehand the keys used in this JSON tree, since the JSON is dynamic, this is not possible.

Tests/MeiliSearchIntegrationTests/UpdatesTests.swift
ppamorim: I will not stack overflow because:

 - There is the intrinsic delayed from the Https requests from SDK and MeiliSearch server.
 - The function `self.wait(for: [expectation], timeout: 10.0)` is async and will fail (and stop) the test when it reaches the limit of 10 seconds running.

This test was designed like this to support async write in the `MeiliSearch` instance while verifying the status. If the test run for more than 10 seconds, it fails and everything stops.

curquiza: But why should we need to check the status here? 
The point of this test is to check we have an update, not to check the update was processed, right?

ppamorim: The test assumes that the input X is received and processed in the Meilisearch instance and the result (in this case `update.status`) from the call `getUpdate` is correctly mapped by the SDK. This test confirms that `status` is being correctly mapped. This will be later transformed into a `enum`.

curquiza (approved):
This PR is finally ready for a merge :)

Two points could be improved from my POV:
- the behemot (see https://github.com/meilisearch/meilisearch-swift/pull/54#discussion_r501550450): I'm going to open an issue
- the if condition (see https://github.com/meilisearch/meilisearch-swift/pull/54#discussion_r500443109). I'm going to open a PR with an improvement suggestion in the next few days.

Thanks again @ppamorim!

curquiza:
bors try

curquiza:
bors merge

PR #71, opened 2020-10-10
Change the update status as enum and rename Type into UpdateType in Update model

ppamorim:
This PR transforms the status result from the `Update` struct into an enum. This change is needed since the server has a stable API and the user doesn't want to keep comparing strings in the SDK.

### DIFF
--- a/Sources/MeiliSearch/Model/Update.swift
+++ b/Sources/MeiliSearch/Model/Update.swift
@@ -55,8 +55,8 @@ public struct Update: Codable, Equatable {
         case processed
         case failed
 
-        public enum CodingError: Error {
-            case unknownStatus
+        public enum StatusError: Error {
+            case unknown
         }
 
         public init(from decoder: Decoder) throws {
@@ -70,21 +70,11 @@ public struct Update: Codable, Equatable {
             case "failed":
                 self = .failed
             default:
-                throw CodingError.unknownStatus
+                throw StatusError.unknown
             }
         }
 
-        public func encode(to encoder: Encoder) throws {
-            var container = encoder.singleValueContainer()
-            switch self {
-            case .enqueued:
-                try container.encode("enqueued")
-            case .processed:
-                try container.encode("processed")
-            case .failed:
-                try container.encode("failed")
-            }
-        }
+        public func encode(to encoder: Encoder) throws { }
 
     }
 
--- a/Sources/MeiliSearch/Model/Update.swift
+++ b/Sources/MeiliSearch/Model/Update.swift
@@ -17,13 +17,13 @@ public struct Update: Codable, Equatable {
         // MARK: Properties
 
         ///Returns if the update has been sucessful or not.
-        public let status: String
+        public let status: Status
 
         ///Unique ID for the current `Update`.
         public let updateId: Int
 
         ///Type of update.
-        public let type: Type
+        public let type: UpdateType
 
         ///Duration of the update process.
         public let duration: TimeInterval?
@@ -34,8 +34,8 @@ public struct Update: Codable, Equatable {
         ///Date when the update has been processed.
         public let processedAt: Date?
 
-        ///Typr of `Update`.
-        public struct `Type`: Codable, Equatable {
+        ///Type of `Update`
+        public struct UpdateType: Codable, Equatable {
 
             // MARK: Properties
 
@@ -45,6 +45,45 @@ public struct Update: Codable, Equatable {
             /// ID of update type.
             public let number: Int
 
+        }
+
+    }
+
+    public enum Status: Codable, Equatable {
+
+        case enqueued
+        case processed
+        case failed
+
+        public enum CodingError: Error {
+            case unknownStatus
+        }
+
+        public init(from decoder: Decoder) throws {
+            let container = try decoder.singleValueContainer()
+            let rawStatus = try container.decode(String.self)
+            switch rawStatus {
+            case "enqueued":
+                self = .enqueued
+            case "processed":
+                self = .processed
+            case "failed":
+                self = .failed
+            default:
+                throw CodingError.unknownStatus
+            }
+        }
+
+        public func encode(to encoder: Encoder) throws {
+            var container = encoder.singleValueContainer()
+            switch self {
+            case .enqueued:
+                try container.encode("enqueued")
+            case .processed:
+                try container.encode("processed")
+            case .failed:
+                try container.encode("failed")
+            }
         }
 
     }
--- a/Tests/MeiliSearchIntegrationTests/UpdatesTests.swift
+++ b/Tests/MeiliSearchIntegrationTests/UpdatesTests.swift
@@ -71,8 +71,8 @@ class UpdatesTests: XCTestCase {
                 self.client.getUpdate(UID: self.uid, update) { result in
 
                     switch result {
-                    case .success(let update):
-                        XCTAssertTrue(["enqueued", "processed", "fail"].contains(update.status))
+                    case .success:
+                        break
                     case .failure(let error):
                         print(error)
                         XCTFail()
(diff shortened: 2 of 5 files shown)


ppamorim:
bors try

Sources/MeiliSearch/Model/Update.swift
```diff
@@ -17,13 +17,13 @@ public struct Update: Codable, Equatable {
...
         ///Returns if the update has been sucessful or not.
-        public let status: String
+        public let status: Status
 
         ///Unique ID for the current `Update`.
         public let updateId: Int
 
         ///Type of update.
-        public let type: Type
+        public let type: ResultType
```
curquiza, line 26:
```suggestion
        public let type: UpdateType
```

Sources/MeiliSearch/Model/Update.swift
```diff
@@ -34,8 +34,8 @@ public struct Update: Codable, Equatable {
         ///Date when the update has been processed.
         public let processedAt: Date?
 
-        ///Typr of `Update`.
-        public struct `Type`: Codable, Equatable {
+        ///Type of `Update`, unfortunately the name `Type` is a Swift type and conflicts with the std value.
```
curquiza, line 37:
```suggestion
        ///Type of `Update`
```

Sources/MeiliSearch/Model/Update.swift
```diff
@@ -34,8 +34,8 @@ public struct Update: Codable, Equatable {
...
+        public struct ResultType: Codable, Equatable {
```
curquiza, line 38:
```suggestion
        public struct UpdateType: Codable, Equatable {
```

Sources/MeiliSearch/Model/Update.swift
curquiza: `Result` is too generic 🙂

Tests/MeiliSearchIntegrationTests/UpdatesTests.swift
```diff
@@ -72,7 +72,7 @@ class UpdatesTests: XCTestCase {
 
                     switch result {
                     case .success(let update):
-                        XCTAssertTrue(["enqueued", "processed", "fail"].contains(update.status))
+                        break
```
curquiza, line 75: Where is the test? 🙂

curquiza:
Great test added! 👍

Tests/MeiliSearchIntegrationTests/UpdatesTests.swift
ppamorim: the test is the type itself, if MeiliSearch returns anything different from the values `ResultType`, the decode will fail. I believe you guys aren't planning to include any new status in the version 0.15, don't you?

curquiza: Nope, but the status names will change into `in_process`, `failed`, and `done`, see https://github.com/meilisearch/MeiliSearch/issues/998.

ppamorim: I will rename it to the new values and add support to the old ones. What about it?

curquiza: Rename them in the future, yes. No need to support the old ones 😉

curquiza:
bors try

ppamorim:
I need to rebase this PR since the test is failing.

ppamorim:
bors try

Tests/MeiliSearchIntegrationTests/UpdatesTests.swift
```diff
@@ -105,11 +105,8 @@ class UpdatesTests: XCTestCase {
         self.client.getAllUpdates(UID: self.uid) { result in
 
             switch result {
-            case .success(let updates):
-                let statuses: [String] = ["enqueued", "processed", "fail"]
-                updates.forEach { (update: Update.Result) in
-                    XCTAssertTrue(statuses.contains(update.status))
-                }
+            case .success:
+                break
```
curquiza, line 109: Sorry to come back to this again! But can't we check at least the type of `let updates`? So that we check the returned value of the method. If someone change it in the future we would be able to see it.

Sources/MeiliSearch/Model/Update.swift
```diff
@@ -49,4 +49,43 @@ public struct Update: Codable, Equatable {
...
+            var container = encoder.singleValueContainer()
+            switch self {
+            case .enqueued:
+                try container.encode("enqueued")
+            case .processed:
+                try container.encode("processed")
+            case .failed:
+                try container.encode("failed")
+            }
+        }
```
bidoubiwa, line 87: Is this method useful for the user? It is used in testing, but other than that I dont see the usage.

Tests/MeiliSearchUnitTests/UpdatesTests.swift
```diff
@@ -46,10 +70,51 @@ class UpdatesTests: XCTestCase {
...
+
+        self.wait(for: [expectation], timeout: 1.0)
+
+    }
+
+    func testGetUpdateInvalidStatus() {
+
+        //Prepare the mock server
+
+        let json = """
```
bidoubiwa, line 87:
```suggestion
        let badStatusUpdate = """
```
Should be changed in the example above aswell but I do not have access with github interface.

Sources/MeiliSearch/Model/Update.swift
```diff
@@ -49,4 +49,43 @@ public struct Update: Codable, Equatable {
 
     }
 
+    public enum Status: Codable, Equatable {
+
+        case enqueued
+        case processed
+        case failed
+
+        public enum CodingError: Error {
```
bidoubiwa, line 58:
```suggestion
        public enum StatusError: Error {
```
What do you think of this less generic name ?

Sources/MeiliSearch/Model/Update.swift
ppamorim: This might not be but since `Status` is a value from `Update.Result` and `Update` that are `Codable`, swift forces it to be Codable too. If I set it as `Decodable`, I get this error:

Screenshot 2020-10-14 at 00 36 53

I removed the implementation.

Sources/MeiliSearch/Model/Update.swift
ppamorim: Replaced it with `StatusError.unknown`.

bidoubiwa (dismissed):
Thanks a lot :)

curquiza (approved):

curquiza:
bors merge

curquiza:
There is clearly a problem when installing MeiliSearch on the CI with brew... I don't have any issues on my local machine. And you guys @eskombro @ppamorim @bidoubiwa?

eskombro:
> ave any issues on my local machine. And you guys @eskombro @ppamorim @bidoubiwa?

I have no issues. Installing meilisearch with brew works fine for me!

We could install it on the CI with `curl -L https://install.meilisearch.com | sh` instead


Maybe we shouldn't rely on brew, since we can download directly the binary fo the specific macOS version

curquiza:
Yes indeed we should change!
The script is done by humans, I would rather use docker as we do in the other SDKs, it works fine:
https://github.com/meilisearch/meilisearch-php/blob/ced74248be12f3ca594ce115419cf28d3ef6d609/.github/workflows/tests.yml#L35-L36

curquiza:
Oh I remember, we spent a lot of time with Charlotte about that! Sorry again Pedro because it's not related to the issue.
We used to use the installation script, but we don't know why, it fails on macos-latest (only on CI, it works locally). See https://github.com/meilisearch/MeiliSearch/issues/908
Docker is not installed on the macos-latest CI.
So, brew was the latest option...

eskombro:
> Let's wait, it's maybe a GitHub issue... It worked a few hours ago.

Ok! Indeed it probably is an authentication issue coming from GitHub. So if it keeps doing the same we can create a new issue and test everything there!

ppamorim:
@eskombro @curquiza Same error here: https://github.com/actions/virtual-environments/issues/1811

ppamorim:
Sorted with this change:

```yml
run: |
    brew uninstall openssl@1.0.2t
    brew uninstall python@2.7.17
    brew untap local/openssl
    brew untap local/python2
    brew update
```

curquiza:
If it's a temporary fix I would rather not add it.
If it's a definitive fix, I would rather have it in another PR please 🙂

curquiza:
bors try

curquiza (approved):

curquiza:
bors merge

Issue #75, opened 2020-10-12
Remove specific indexAlreadyExist error.


bidoubiwa:
## Problem

There is a class called `CreateError` that created an error called `indexAlreadyExists`.

This is used in the `getOrCreateIndex` method to determine if an index already exists: 
```swift

    func getOrCreate(_ UID: String, _ completion: @escaping (Result<Index, Swift.Error>) -> Void) {
        self.create(UID) { result in
            switch result {
            case .success(let index):
                completion(.success(index))

            case .failure(let error):
                switch error {
                case CreateError.indexAlreadyExists: // "index already exists" should be used using error.errorCode
                    self.get(UID, completion)
                default:
                    completion(.failure(error))
                }
            }
        }
    }
```

```swift
/**
Error type for all functions included in the `Indexes` struct.
*/
public enum CreateError: Swift.Error, Equatable {

    // MARK: Values

    /// Case the `Index` already exists in the Meilisearch instance, this error will return.
    case indexAlreadyExists

    // MARK: Codable

    static func decode(_ error: MSError) -> Swift.Error {
        let underlyingError: NSError = error.underlying as NSError
        if let msErrorResponse: MSErrorResponse = error.data {
            if underlyingError.code == 400 && msErrorResponse.errorType == "invalid_request_error" && msErrorResponse.errorCode == "index_already_exists" {
                return CreateError.indexAlreadyExists
            }
            return error

        }
        return error
    }
}
```

## Solution

We should be able to do something like this without having to go through another error class.
```swift
    func getOrCreate(_ UID: String, _ completion: @escaping (Result<Index, Swift.Error>) -> Void) {
        self.create(UID) { result in
            switch result {
            case .success(let index):
                completion(.success(index))
            case .failure(let error):
                switch error {
                case error.errorCode == "index_already_exists":
                    self.get(UID, completion)
                default:
                    completion(.failure(error))
                }
            }
        }
    }
```

bidoubiwa:
Fixed by: #145

Issue #40, opened 2020-08-11
Create a better error handler


bidoubiwa:
When doing tests the only error I recieve whenever there is an error is the following:

```
failed - The operation couldn’t be completed. (MeiliSearch.MSError error 1.)
```

I have to look at meiliSearch logs to understand the problem.

The error handler should return the full body that meiliSearch provides.

## More precision (by @curquiza)

We should add a basic error handler with these 2 custom exception (or error):
- `MeiliSearchApiException`: MeiliSearch (HTTP) sends back a 4xx or a 5xx. There are most of the SDKs errors.
- `MeiliSearchCommunicationException`: problem with communication, for example, the MeiliSearch instance is unreachable. 

FYI: Even if it's not documented yet, MeiliSearch API has a great error handler, and returns this kind of JSON when an error is found:

<img width="603" alt="Capture d’écran 2020-07-02 à 10 18 21" src="https://user-images.githubusercontent.com/20380692/86334372-6880da00-bc4d-11ea-84ae-82b090394923.png">

It could be great to let the user access `message`, `errorCode`, `errorType` and `errorLink`.

ppamorim:
That's because the current tests where using `XCTFail(error.localizedDescription)` to print the error. This is wrong since not all the error is print by this function. I did replace this with `print(error)` in the PR #54.

bidoubiwa:
fixed by: #145

PR #76, opened 2020-10-12
Create error handler

bidoubiwa:
Addresses the following issues: #75 and #40 

## Problem need addressing 

Cannot make error with standard Error body.

----

In this PR I try to transform MeiliSearch answer in an `MSerror` type.

Tests are failing, I would like some advice to how to make them work (see comments inside code).

Also I would like to remove the whole `CreateError` class. Which I'm not sure is really useful as it removed the purposed of having an `msError` object from which I should be able to look through every data MeilISearch send me:

- msError.errorCode
- msError.errorLink
- msError.errorCode 
- msError.message
- msError.errorType

Output print: 
```
MSError(data: Optional(MeiliSearch.MSErrorResponse(message: "Document with id 123456 not found", errorCode: "document_not_found", errorType: "invalid_request_error", errorLink: Optional("https://docs.meilisearch.com/errors#document_not_found"))), underlying: Error Domain=HttpStatus Code=404 "(null)")
```

##  TODO 
- [ ] refactor request by putting the error handler in a dedicated file and not 4 copies of it in requests
- [ ] Remove `createError`
- [ ] Fix error that does not accept standard body error

### DIFF
--- a/Sources/MeiliSearch/Indexes.swift
+++ b/Sources/MeiliSearch/Indexes.swift
@@ -70,8 +70,9 @@ struct Indexes {
                 completion(.success(index))
 
             case .failure(let error):
+                print(error)
                 switch error {
-                case CreateError.indexAlreadyExists:
+                case CreateError.indexAlreadyExists: // "index already exists" should be used using error.errorCode
                     self.get(UID, completion)
                 default:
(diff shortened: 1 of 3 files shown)


bidoubiwa:
fixed with #145

Issue #94, opened 2020-12-20
Cannot change default SearchParameters


felix-gohla:
# Problem

After looking through the documentation and the code, I did not find any way to change the default `SearchParameters`.
There is a static function for creating a `query` with default parameters, but I do not know how to e.g. query with a specific offset.
Am I missing something here? 😊 

# Solution

One easy fix was to make the initialiser of `SearchParameters` public. If this is not desired, the fields could be variables instead of constants, so you could change the fields after using the `query` function.

curquiza:
Hello @felix-gohla,

Is this test an example of what you are looking for? https://github.com/meilisearch/meilisearch-swift/blob/0af420d4ed40f6fd669dd80de24c4d429ddcdbc0/Tests/MeiliSearchIntegrationTests/SearchTests.swift#L288-L311
You can set a specific `offset` for your search.

Sorry if I misunderstood your request!

felix-gohla:
Hi @curquiza,

thank you very much for responding that quickly. 😊 
Yes, this test shows exactly what I am looking for, but since I am using this library in another project, the initialiser is not exposed. The test that you linked has access to the initialiser because it can access the package's internal functions, but it is not `public`. Therefore, I am not able to create a new `SearchParameters` object using the standard initialiser (see the attached screenshot) from outside the library.

<img width="665" alt="Screenshot 2020-12-21 at 11 56 14" src="https://user-images.githubusercontent.com/37421906/102770135-efed1180-4383-11eb-94ea-152e1daa3e40.png">

curquiza:
Ok! I understand better! Do you want to open a PR to fix this since you seem to exactly know how to fix it? Thus, you will appear as the contributors of this repo 🙂 (you need to specify your email address in your commits).
See our contribution guidelines: https://github.com/meilisearch/meilisearch-swift/blob/master/CONTRIBUTING.md#how-to-contribute
If you don't have the time I will do it!

felix-gohla:
Sorry, didn't have time the last two days, but I will create one! 👍

curquiza:
closed by #98

PR #98, opened 2020-12-23
Make SearchParameters initializer public

felix-gohla:
Addresses Issue: #94 

# Summary

This fixes an issue where the `SearchParameters` could be modified from outside the library.
By making the initializer public, the `SearchParameters` can be configured individually.

# Discussion

I am not sure whether tests are necessary in this case, as this is a very simple fix changing only one line.


bidoubiwa (approved):
Since tests have access to the internal function, I'm not sure there is a way to tests if it is also publicly available.

Thanks a lot for this PR! It is a very useful addition.

bidoubiwa:
bors merge

PR #113, opened 2021-03-24
Fix health tests

curquiza:


Tests/MeiliSearchUnitTests/SystemTests.swift
```diff
@@ -16,7 +16,13 @@ class SystemTests: XCTestCase {
...
         //Prepare the mock server
 
-        session.pushEmpty(code: 204)
+        let jsonString = """
+        {
+            "status": "available"
+        }
+        """
+
+        session.pushData(jsonString, code: 200)
```
bidoubiwa, line 25: JsonString should be parsed in Swift and compared with the returned body: 

```swift
       let jsonString = """
        {
            "status": "available"
        }
        """


        session.pushData(jsonString, code: 200)
       
        let jsonData = jsonString.data(using: .utf8)!

        let expectedHealthBody: Version = try! Constants.customJSONDecoder.decode(Version.self, from: jsonData)

        // Start the test with the mocked server
        let expectation = XCTestExpectation(description: "Check server health")

        self.client.health { result in
            switch result {
            case .success(let body):
                XCTAssertEqual(expectedHealthBody, body)
                expectation.fulfill()
            case .failure:
                XCTFail("Failed to check server health")
            }
        }

        self.wait(for: [expectation], timeout: 1.0)
```

I'm not sure why we left the sessions, maybe @ppamorim knows?

bidoubiwa (changes requested):

Tests/MeiliSearchUnitTests/SystemTests.swift
curquiza: I had to add a `Health` struct. Please check 🙂

bidoubiwa (approved):
Very nice struct! 

LGTM 🔥

curquiza:
bors try

PR #115, opened 2021-04-21
Standardize health method

alallema:
**Description**
Checking that method health() return {'status': 'available'} and added isHealthy() method who return boolean value.

**Issue related**
meilisearch/integration-guides#55.

### DIFF
--- a/Tests/MeiliSearchUnitTests/SystemTests.swift
+++ b/Tests/MeiliSearchUnitTests/SystemTests.swift
@@ -12,7 +12,7 @@ class SystemTests: XCTestCase {
         client = try! MeiliSearch(Config(hostURL: nil, session: session))
     }
 
-    func testHealth() {
+    func testHealthStatusAvailable() {
 
         //Prepare the mock server
 
@@ -30,7 +30,7 @@ class SystemTests: XCTestCase {
 
         // Start the test with the mocked server
 
-        let expectation = XCTestExpectation(description: "Check server health")
+        let expectation = XCTestExpectation(description: "Check body of health server on health method")
 
         self.client.health { result in
             switch result {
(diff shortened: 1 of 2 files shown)


Tests/MeiliSearchUnitTests/SystemTests.swift
```diff
@@ -46,6 +46,50 @@ class SystemTests: XCTestCase {
...
+        session.pushData("", code: 400)
+
+        // Start the test with the mocked server
+
+        let expectation = XCTestExpectation(description: "Check server health")
+
+        self.client.isHealthy { result in
+            XCTAssertEqual(result, false)
+            expectation.fulfill()
+        }
```
bidoubiwa, line 87: I see you'r not catching a possible error: 

Like the following example
```swift
     switch result {
                case .success:
                    expectation.fulfill()
                case .failure(let error):
                    print(error)
                    XCTFail()
                }
            }
```

What happens when `testIsHealthyFail` fails? Is the output correct?

alallema: You right I fix it!

Tests/MeiliSearchUnitTests/SystemTests.swift
```diff
@@ -30,15 +30,67 @@ class SystemTests: XCTestCase {
...
+
+        //Prepare the mock server
+
+        let jsonString = """
+        {
+            "status": "available"
+        }
+        """
+
+        session.pushData(jsonString, code: 200)
```
bidoubiwa, line 59:
```suggestion

        session.pushData("", code: 200)
```

bidoubiwa (changes requested):

Tests/MeiliSearchUnitTests/SystemTests.swift
alallema: The test failed without the json. I think it's because the `isHealthy()`  method call `health()` method who need it.

bidoubiwa: Oh yes probably! I forgot about that :)

bidoubiwa (approved):
bors merge

Issue #133, opened 2021-05-13
Instantiate the MeiliSearch client


curquiza:
The `MeiliSearch` method that instantiates the client:
- [ ] should take a mandatory parameter: the host URL
- [ ] should not do an HTTP call to the server

Issue #97, opened 2020-12-21
Remove linter warnings


curquiza:
There is some linter warning we can see at each PR: https://github.com/meilisearch/meilisearch-swift/pull/96/files

This issue is to make them desappear.

Please, feel free to ask any question if not clear 🙂

ppamorim:
Yes I need to remove the linter checks for the unit tests. I will do it soon.

PR #134, opened 2021-05-19
Fix all swiftlint errors and remove Tests from the linter

ppamorim:
Hi,

In this PR I fixed all swiftlint errors and removed the Tests and Demos folder for the linter, this sorts the issue #97.
I did fix the errors in points of the code that are going to be removed, as listed on #133.

My local linter is on the latest version 0.42.0.

Multiple linter fixes applied by running `swiftlint autocorrect`.

Regards.


ppamorim:
Any idea why the checks are being cancelled? Internal error?

ppamorim:
@curquiza I had to update Xcode to 12.4 to support `decode(String?.self, ...)` otherwise it crashes on Xcode 11.5.

.swiftlint.yml
```diff
@@ -23,8 +23,9 @@ opt_in_rules: # some rules are only opt-in
 excluded:
   - Carthage
   - Pods
-  - VaporDemo/.build
-  - PerfectDemo/.build
+  - Demos/VaporDemo
+  - Demos/PerfectDemo
+  - Tests
```
bidoubiwa, line 28: Why do we exclude these folders?

Sources/MeiliSearch/Config.swift
```diff
@@ -11,8 +11,8 @@ public class Config {
 
     // MARK: Static
 
-    ///Deafault configuration for the default MeiliSearch host, do not use this in
-    ///production since it does not contains the apiKey.
+    /// Deafault configuration for the default MeiliSearch host, do not use this in
```
bidoubiwa, line 14:
```suggestion
    /// Default configuration for the default MeiliSearch host, do not use this in
```
( nothing to do with your PR but why not hihi)

bidoubiwa (changes requested):

bidoubiwa:
Thanks a lot for this PR! Sorry for the delay

.swiftlint.yml
ppamorim: We don't need swiftlint in the demos and tests folder at the moment. I can open a new PR to sort them all later and reintroduce them back to the swiftlint.

bidoubiwa: Hey, really sorry @ppamorim but linting is bound to a project as a whole not only the source code. It creates consistency and makes contributing easier. 
It would be weird if I don't have to follow the same rules in the tests folder than in the source folder.

bidoubiwa (changes requested):
~per the discussion we have in a [thread](https://github.com/meilisearch/meilisearch-swift/pull/134/files#r642896433).~

~For consistency in coding practices, I prefer linting to be applied in tests and demo's dir's as well.~

bidoubiwa:
bors merge

PR #150, opened 2021-06-29
Add SwiftLint for demos and tests

ppamorim:
In this PR the demos and tests are back in the SwiftLint. I tried to exclude the files in the `.swiftlint.yml` file but it did work and resulted in a more complex solution than disabling each broken and acceptable rule in the source code by using `swiftlint:disable rule`.

Rules disabled:

 - `force_try`: During tests it's normal to expect non try values and force unwrapping the initialisations and conversions of data, if it's crashing, there is something wrong with the library or the test is wrong. On the examples, it's Ok to crash the test or demo on the initialisation of the client, but the force conversion of data has been removed to a safe alternative.
 - `force_unwrapping`: Similar as `force_try`.
 - `force_cast`: There are some objects that are being converted to objc values (like `NSError` (or `JSONSerialization` in other tests)) in the `ClientsTests`, converting them with force cast is a part of the test to verify the the expected value matches with the tested value.
 - `line_length`: The `DocumentsTests` unit test is very long, well over 200 lines due the number of spaces, it could be massively reduced but I am afraid of changing it because it's well... not that big, being only 424. I would like to keep this file as an exception to the rule because it's a test, it could be different if it was in the library source code.

Please let me know if this requires any change.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -3,8 +3,13 @@ disabled_rules:
   - type_body_length
   - file_length
   - nesting # should be re-enabled
+  - xctfail_message
 
-opt_in_rules: # some rules are only opt-in
+analyzer_rules:
+  - unused_declaration
+  - unused_import
+
+opt_in_rules:
   - explicit_type_interface
   - force_unwrapping
   - implicit_return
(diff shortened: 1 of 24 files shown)


bidoubiwa:
bors try

bidoubiwa:
Could you rebase @ppamorim ?

ppamorim:
@bidoubiwa Done :)

ppamorim:
@bidoubiwa Could you please review it again?

bidoubiwa (approved):
LGTM 🔥

bidoubiwa:
bors merge

PR #152, opened 2021-06-30
Fix issue where settings are being erased and add tests

ppamorim:
Hi,

In this PR I fixed the issue where the settings are being deleted, this was happening because the library didn't know the different between the delete, removal and keep of a value. In my solution I had to add the `Action` enum (should I rename it to `Transaction`?) that flags the type of `Action` and a  new struct only used on the update settings. Reusing the same `Setting` struct could make the read from Meilisearch server more complex because everything is wrapped around `Action`.

Unfortunately these are breaking changes. If you have a better solution, please let me know. We could transform all arrays to optional arrays, but it could be a problem with the `distinctAttribute` which is already an optional value.

Regards,
Pedro

### DIFF
--- a/Sources/MeiliSearch/Client.swift
+++ b/Sources/MeiliSearch/Client.swift
@@ -418,7 +418,7 @@ public struct MeiliSearch {
      */
     public func updateSetting(
         UID: String,
-        _ setting: Setting,
+        _ setting: UpdateSetting,
         _ completion: @escaping (Result<Update, Swift.Error>) -> Void) {
         self.settings.update(UID, setting, completion)
     }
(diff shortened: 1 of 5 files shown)


bidoubiwa:
The usage should be made more simple. An investigation should be done to find alternative solutions. 

When keys are missing they should not be added to the body sent to MeiliSearch.

bidoubiwa (changes requested):
The usage should be made more simple. An investigation should be done to find alternative solutions.

When keys are missing they should not be added to the body sent to MeiliSearch.

bidoubiwa:
Thanks so much for your contribution @ppamorim 
I found a working solution at #188 

fixed by: #188

PR #190, opened 2021-09-14
Change get/update/resetSetting to get/update/resetSettings with a s

bidoubiwa:
For consistency between SDK the naming of the following methods have been changed: 

- `getSetting` becomes `getSettings`
- `updateSetting` becomes `updateSettings`
- `resetSetting` becomes `resetResstings`

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -153,7 +153,7 @@ get_keys_1: |-
       }
   }
 get_settings_1: |-
-  client.getSetting(UID: "movies") { (result: Result<Setting, Swift.Error>) in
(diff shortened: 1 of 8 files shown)


curquiza (commented):
In the code samples and in SettingsTests.swift I still see some variable at singular (`Setting`). Don't know if it's expected

bidoubiwa:
@curquiza yes this is normal! It is not linked to the `Settings` that is the class, but the `Setting` that is the model of the settings. 

See: https://github.com/meilisearch/meilisearch-swift/blob/main/Sources/MeiliSearch/Model/Setting.swift

alallema (approved):
LGTM!

bidoubiwa:
bors merge

PR #200, opened 2021-09-21
Move stats method to index class

bidoubiwa:


Sources/MeiliSearch/Client.swift
```diff
@@ -36,17 +35,15 @@ public struct MeiliSearch {
   public init(host: String, apiKey: String? = nil, session: URLSessionProtocol? = nil) throws {
     self.config = try Config(host: host, apiKey: apiKey, session: session).validate()
     let request: Request = Request(self.config)
-    // self.indexes = Indexes(self.config)
     self.keys = Keys(request, self.config)
-    self.settings = Settings(request)
     self.stats = Stats(request)
```
alallema, line 39: And here?

Sources/MeiliSearch/Client.swift
```diff
@@ -19,7 +19,6 @@ public struct MeiliSearch {
 
   // private let indexes: Indexes
   private let keys: Keys
-  private let settings: Settings
   private let stats: Stats
```
alallema, line 22: Do we remove it here too?

alallema: Not settings but stats

Sources/MeiliSearch/Client.swift
alallema: Not settings but stats

PR #205, opened 2021-10-04
Fix: remove charset=utf-8 from Content-Type header

curquiza:


bidoubiwa (commented):
I see you didn't setValue for the `get` method. Is there a reason?

curquiza:
> I see you didn't setValue for the get method. Is there a reason?

Which part do you mean? There is no `setValue` regarding the content type for the `get` method, right?

bidoubiwa:
Ah yes, my bad!

bidoubiwa (approved):
bors merge

PR #232, opened 2021-11-11
Add a new function deleteIfExists

brunoocasali:
## What does this PR do?
Fixes https://github.com/meilisearch/meilisearch-swift/issues/208

## PR checklist
- [x] Does this PR fix an existing issue?

Thank you so much for contributing to MeiliSearch!

### DIFF
--- a/Tests/MeiliSearchUnitTests/IndexesTests.swift
+++ b/Tests/MeiliSearchUnitTests/IndexesTests.swift
@@ -277,6 +277,45 @@ class IndexesTests: XCTestCase {
     self.wait(for: [expectation], timeout: 10.0)
   }
 
+  func testDeleteIndexIfExists() {
+    // Prepare the mock server
+    session.pushEmpty(code: 204)
+
+    // Start the test with the mocked server
+    let expectation = XCTestExpectation(description: "Delete Movies index")
+
+    self.index.deleteIfExists { result in
+      if result {
+        XCTAssertTrue(result)
+        expectation.fulfill()
+      } else {
+        XCTFail("Failed to delete Movies index, it was not present on the server")
+      }
+      expectation.fulfill()
+    }
+
+    self.wait(for: [expectation], timeout: 10.0)
+  }
+
+  func testDeleteIndexIfExistsWhenIsnt() {
+    // Prepare the mock server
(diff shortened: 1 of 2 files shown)


Tests/MeiliSearchUnitTests/IndexesTests.swift
```diff
@@ -277,6 +277,45 @@ class IndexesTests: XCTestCase {
...
+
+    // Start the test with the mocked server
+    let expectation = XCTestExpectation(description: "Delete Movies index only if exists")
+
+    self.index.deleteIfExists { result in
+      if !result {
+        XCTAssertFalse(result)
+        expectation.fulfill()
+      } else {
+        XCTFail("Failed to delete Movies index, it was not present on the server")
```
bidoubiwa, line 312:
```suggestion
        XCTFail("Deleting the index should have returned false as the index does not exist on the server")
```

bidoubiwa (changes requested):
Thanks so much for this PR 🔥 It was greatly needed

Tests/MeiliSearchUnitTests/IndexesTests.swift
brunoocasali: @bidoubiwa thanks for your comment, I tried to find something like this https://betterspecs.org/ for swift but unfortunately I only found some random articles about this. Do you have something like that to share?

bidoubiwa: @brunoocasali  this is a very interesting ressource. I'm not the best at writing good and consistent specs. I will definitely give it a look as it would mean to change very spec description of the repository

brunoocasali: Oh, actually it was not a criticism ok? It's because in ruby community we have "guides" everywhere haha an I like to follow them to evict "bikeshedding", but and I really liked what you did on this project :)

bidoubiwa: Oh no don't worry, I totally agree that we should follow some specs best practices! I didn't take it as a criticism but as a good resource that I want to follow :)

bidoubiwa (approved):
bors merge

Issue #235, opened 2021-11-12
📐 Adopt a stricter linter ruler (raywenderlich)


brunoocasali:
It seems that https://github.com/raywenderlich/swift-style-guide is a linting set of rules that the swift community adopted.

Running locally in it found `Done linting! Found 894 violations, 6 serious in 54 files.`
The proposal is to adopt this style guide in 5 steps:

- [x] Open the PR with the configuration (but as a draft, it should be merged only after step 4) https://github.com/meilisearch/meilisearch-swift/pull/236
- [x] 1. Add the new configuration to CI  https://github.com/meilisearch/meilisearch-swift/pull/236
- [x] 2. Fix automatically correctable offenses https://github.com/meilisearch/meilisearch-swift/pull/237
- [ ] 3. Fix other small offenses (one kind per PR?)
- [ ] 4. Ignore the error offenses and create separated issues for them

brunoocasali:
Working in this vast refactor in the following months will not be possible. Since I'm cleaning the repo, I will close this.

PR #236, opened 2021-11-12
Add stricter configuration rule for the project

brunoocasali:
https://github.com/raywenderlich/swift-style-guide

Related to #235

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,43 +1,109 @@
-disabled_rules:
-  - todo
-  - type_body_length
-  - file_length
-  - nesting # should be re-enabled
-  - xctfail_message # should be re-enabled
-  - identifier_name
-  - function_body_length
+excluded:
+  - ${PWD}/Carthage
+  - ${PWD}/Pods
+  - ${PWD}/DerivedData
+  - ${PWD}/.build
+  - ${PWD}/Demos/PerfectDemo/.build
+  - ${PWD}/Demos/VaporDemo/.build
 
-analyzer_rules:
-  - unused_declaration
-  - unused_import
+disabled_rules:
+  - discarded_notification_center_observer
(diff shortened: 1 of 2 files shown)


com.raywenderlich.swiftlint.yml
```diff
@@ -0,0 +1,109 @@
+excluded:
+  - ${PWD}/Carthage
+  - ${PWD}/Pods
+  - ${PWD}/DerivedData
+  - ${PWD}/.build
+  - ${PWD}/Demos/PerfectDemo/.build
```
brunoocasali, line 6: Unfortunately the swiftlint do not support wildcards for this use case :/

.swiftlint.yml
```diff
@@ -1,43 +1,109 @@
+excluded:
+  - ${PWD}/Carthage
+  - ${PWD}/Pods
+  - ${PWD}/DerivedData
+  - ${PWD}/.build
+  - ${PWD}/Demos/PerfectDemo/.build
+  - ${PWD}/Demos/VaporDemo/.build
```
brunoocasali, line 7: Unfortunately the swiftlint do not support wildcards for this use case :/

bidoubiwa: Do you mean we have to specify file by file ?

brunoocasali: Not the "file" itself, but the directory, for example for `.build` we can't just use: ` - ${PWD}/**/.build/"` :(

brunoocasali:
> Hey @brunoocasali is this still relevant?

Its is something that makes our SDK better, but this is something we can live without, so is up to you, I would like to continue working on this issue in the future :shipit:

brunoocasali:
Unfortunately, I don't have time to work on this code anymore. Closing!

PR #256, opened 2022-02-08
Remove GetOrCreateIndex and DeleteIfExistsIndex

bidoubiwa:
## Breaking

- `getOrCreateIndex` no longer exists. Create index returns a Task as index creation became asynchronous.
- `deleteIfExistsIndex` no longer exists. Because deleting an index is now an asynchronous task it will either delete the index or the task will fail. But it will not block any other action.


brunoocasali (approved):
LGTM! 🦅

PR #267, opened 2022-02-14
Update version update in contributing guide

bidoubiwa:
The [check-release](https://github.com/meilisearch/meilisearch-swift/blob/main/.github/scripts/check-release.sh) tests ensures that all the mention of the versions of the swift package have been updated before a release. 

Per the contributing it only mentioned one place where to change it. This PR updates with all the lines that must be updated


bidoubiwa:
bors merge

PR #285, opened 2022-04-26
Fix tests dependent on a specific return order for the documents

bidoubiwa:
As per https://github.com/meilisearch/meilisearch/pull/2298 the default returned document order when no relevancy rules are applied has been changed.

### DIFF
--- a/Tests/MeiliSearchIntegrationTests/DocumentsTests.swift
+++ b/Tests/MeiliSearchIntegrationTests/DocumentsTests.swift
@@ -148,8 +148,8 @@ class DocumentsTests: XCTestCase {
               case .success(let returnedMovies):
                 let returnedMovie = returnedMovies[0]
                 XCTAssertEqual(returnedMovies.count, 1)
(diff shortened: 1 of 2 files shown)


Tests/MeiliSearchIntegrationTests/DocumentsTests.swift
```diff
@@ -148,8 +148,8 @@ class DocumentsTests: XCTestCase {
               case .success(let returnedMovies):
                 let returnedMovie = returnedMovies[0]
                 XCTAssertEqual(returnedMovies.count, 1)
-                XCTAssertEqual(returnedMovie.id, 123)
-                XCTAssertEqual(returnedMovie.title, "Pride and Prejudice")
+                XCTAssertEqual(returnedMovie.id, 456)
+                XCTAssertEqual(returnedMovie.title, "Le Petit Prince")
```
alallema, line 152: Do you know why the results changes?

PR #288, opened 2022-04-26
Add new search parameters highlightPreTag, highlightPostTag and cropMarker

bidoubiwa:
As per https://github.com/meilisearch/meilisearch/issues/2214 new search parameters are introduced: 

- `highlightPreTag`
- `highlightPostTag`
- `cropMarker`

### DIFF
--- a/Sources/MeiliSearch/Model/SearchParameters.swift
+++ b/Sources/MeiliSearch/Model/SearchParameters.swift
@@ -26,8 +26,17 @@ public struct SearchParameters: Codable, Equatable {
   /// Limit length at which to crop specified attributes.
   public let cropLength: Int?
 
+  /// Marker in front and behind a cropped value.
+  public let cropMarker: String?
+
   /// Which attributes to highlight.
   public let attributesToHighlight: [String]?
+
+  /// Tag in front of highlighted term(s).
+  public let highlightPreTag: String?
+
+  /// Tag at the end of highlighted term(s).
+  public let highlightPostTag: String?
 
   /// Filter on attributes values.
   public let filter: String?
@@ -50,7 +59,10 @@ public struct SearchParameters: Codable, Equatable {
     attributesToRetrieve: [String]? = nil,
     attributesToCrop: [String]? = nil,
     cropLength: Int? = nil,
+    cropMarker: String? = nil,
     attributesToHighlight: [String]? = nil,
+    highlightPreTag: String? = nil,
+    highlightPostTag: String? = nil,
     filter: String? = nil,
     sort: [String]? = nil,
     facetsDistribution: [String]? = nil,
@@ -61,7 +73,10 @@ public struct SearchParameters: Codable, Equatable {
     self.attributesToRetrieve = attributesToRetrieve
     self.attributesToCrop = attributesToCrop
     self.cropLength = cropLength
+    self.cropMarker = cropMarker
     self.attributesToHighlight = attributesToHighlight
+    self.highlightPreTag = highlightPreTag
+    self.highlightPostTag = highlightPostTag
     self.filter = filter
     self.sort = sort
     self.facetsDistribution = facetsDistribution
@@ -89,7 +104,10 @@ public struct SearchParameters: Codable, Equatable {
     case attributesToRetrieve
     case attributesToCrop
     case cropLength
+    case cropMarker
     case attributesToHighlight
+    case highlightPreTag
+    case highlightPostTag
     case filter
     case sort
     case facetsDistribution
--- a/Tests/MeiliSearchIntegrationTests/SearchTests.swift
+++ b/Tests/MeiliSearchIntegrationTests/SearchTests.swift
@@ -338,6 +338,34 @@ class SearchTests: XCTestCase {
     self.wait(for: [expectation], timeout: TESTS_TIME_OUT)
   }
 
+   // MARK: Crop Marker
+
+  func testSearchCropMarker() {
+    let expectation = XCTestExpectation(description: "Search for Books with a custom crop marker")
+
+    typealias MeiliResult = Result<SearchResult<Book>, Swift.Error>
+    let query = "Manuel"
+    let attributesToCrop = ["comment"]
+    let cropLength = 2
+    let cropMarker = "(ꈍᴗꈍ)"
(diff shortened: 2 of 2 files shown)


Tests/MeiliSearchIntegrationTests/SearchTests.swift
```diff
@@ -338,6 +338,34 @@ class SearchTests: XCTestCase {
...
+    let searchParameters = SearchParameters(query: query, attributesToCrop: attributesToCrop, cropLength: cropLength, cropMarker: cropMarker)
+
+    self.index.search(searchParameters) { (result: MeiliResult) in
+      switch result {
+      case .success(let documents):
+        let book: Book = documents.hits[0]
+        XCTAssertEqual("(ꈍᴗꈍ)Joaquim Manuel(ꈍᴗꈍ)", book.formatted!.comment!)
+        expectation.fulfill()
+      case .failure(let error):
+        print(error)
```
brunoocasali, line 360: WDYT to use `dump(error)` in every `case .failure(let error):`?

Tests/MeiliSearchIntegrationTests/SearchTests.swift
```diff
@@ -438,6 +466,34 @@ class SearchTests: XCTestCase {
...
+  // MARK: Attributes to highlight
+
+  func testSearchPrePostHighlightTags() {
+    let expectation = XCTestExpectation(description: "Search for Books using custom pre and post highlight tags")
+
+    typealias MeiliResult = Result<SearchResult<Book>, Swift.Error>
+    let query = "Joaquim Manuel de Macedo"
+    let attributesToHighlight = ["comment"]
+    let highlightPreTag = "(⊃｡•́‿•̀｡)⊃ "
+    let highlightPostTag = " ⊂(´• ω •`⊂)"
```
brunoocasali, line 478: Asking myself at midnight, what happens if we pass `highlightPostTag = ""`?

Tests/MeiliSearchIntegrationTests/SearchTests.swift
```diff
@@ -438,6 +466,34 @@ class SearchTests: XCTestCase {
...
+    let parameters = SearchParameters(query: query, attributesToHighlight: attributesToHighlight, highlightPreTag: highlightPreTag, highlightPostTag: highlightPostTag)
+
+    self.index.search(parameters) { (result: MeiliResult) in
+      switch result {
+      case .success(let documents):
+        let book = documents.hits[0]
+        XCTAssertTrue(book.formatted!.comment!.contains("(⊃｡•́‿•̀｡)⊃ Joaquim ⊂(´• ω •`⊂) (⊃｡•́‿•̀｡)⊃ Manuel ⊂(´• ω •`⊂) (⊃｡•́‿•̀｡)⊃ de ⊂(´• ω •`⊂) (⊃｡•́‿•̀｡)⊃ Macedo ⊂(´• ω •`⊂)"))
```
brunoocasali, line 485: Can you remove this nullable checks as well?

Tests/MeiliSearchIntegrationTests/SearchTests.swift
bidoubiwa: It is a good idea!

bidoubiwa: #293

Sources/MeiliSearch/Model/SearchParameters.swift
```diff
@@ -26,9 +26,18 @@ public struct SearchParameters: Codable, Equatable {
   /// Limit length at which to crop specified attributes.
   public let cropLength: Int?
 
+  /// Marker in front and behind a cropped value.
```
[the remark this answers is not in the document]
bidoubiwa:
```suggestion
  /// Marker that wraps a cropped value.
```

Tests/MeiliSearchIntegrationTests/SearchTests.swift
```diff
@@ -338,6 +338,34 @@ class SearchTests: XCTestCase {
...
+    let attributesToCrop = ["comment"]
+    let cropLength = 2
+    let cropMarker = "(ꈍᴗꈍ)"
+    let searchParameters = SearchParameters(query: query, attributesToCrop: attributesToCrop, cropLength: cropLength, cropMarker: cropMarker)
+
+    self.index.search(searchParameters) { (result: MeiliResult) in
+      switch result {
+      case .success(let documents):
+        let book: Book = documents.hits[0]
+        XCTAssertEqual("(ꈍᴗꈍ)Joaquim Manuel(ꈍᴗꈍ)", book.formatted!.comment!)
```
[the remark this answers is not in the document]
bidoubiwa: >  error: value of optional type 'FormattedBook?' must be unwrapped to refer to member 'comment' of wrapped base type 'FormattedBook'

I could unwrap it more elegantly but since these are tests I'm not sure it is worth the effort.

Tests/MeiliSearchIntegrationTests/SearchTests.swift
bidoubiwa: Nothing happens! It doesnt add any character behind the word.

Tests/MeiliSearchIntegrationTests/SearchTests.swift
bidoubiwa: See above! It needs to be unwrapped either with `!` or by doing if/guard conditions.

brunoocasali (approved):
🍗 Aweesome!

Tests/MeiliSearchIntegrationTests/SearchTests.swift
bidoubiwa: I think this is a design choice that would go against the native API's behavior. Nonetheless maybe we should suggest it in the product. 
Now, for my opinion. I think we should leave the possibility of an empty string. If the user wants to try things out, temporarily create an empty string, create it in a specific process or any other situation. If the user add an empty string and does not understand why no tags are added, I think the first thing he will check is its search parameters

Tests/MeiliSearchIntegrationTests/SearchTests.swift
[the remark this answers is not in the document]
brunoocasali: I agree, I said so, because it is a configuration from our linter!

PR #277, opened 2022-03-30
Feature/Analytics

brunoocasali:
- Add `User-Agent` inside the pre-defined headers

Add Swift support as requested here meilisearch/integration-guides#150

PR #276, opened 2022-03-29
Feature/Add Qualified version

brunoocasali:
- Add a linter to prevent cocoapods errors before publishing the package.
- Add a dynamic version during the cocoapod publishing.
- Update the publishing script to protect tag creation without updating the `PackageVersion.swift`

PR #295, opened 2022-05-05
Fix releaser script

brunoocasali:
After the implementation of the #276 and #277. 
The `Meilisearch.podspec` does not need to be updated new release, which makes the checking useless.


brunoocasali:
To the reviewers, check commit by commit it will help to understand the diff :)

Issue #296, opened 2022-05-08
generateTenantToken issue related to the JWTKit dependency


brunoocasali:
I made a huge mistake when I choose `vapor/jwt-kit` to implement the generate tenant token method changes from Meilisearch version v0.26. It was apparently a good choice, but I discover the problem after [trying to publish it](https://github.com/meilisearch/meilisearch-swift/runs/6311703826?check_suite_focus=true). 

Vapor JWTKit seemed to be the best choice out there because it had a simple and powerful protocol API and is still maintainable if we compare this package with others like: https://github.com/kylef/JSONWebToken.swift (they seem to be abandoned). 

## Problem:

Vapor `JWTKit` has only support for SPM and has **no support for CocoaPods** (they even have an open issue https://github.com/vapor/jwt-kit/issues/47).

The major issue I'm facing now is what we should do now:

- Remove JWTKit and do the token generation by hand?
- Stop publishing to CocoaPods?
- Remove JWTKit and find another open-source solution out there?
- Fork Vapor JWTKit and add the support for CocoaPods ourselves?

## Planned actions:

- [x] Open this issue explaining the whole situation.
- [ ] Add the `pod spec lint` into the Github Action CI to prevent problems like this from happening again.
- [ ] Remove/Revert the code responsible to do the tenant token generation.
- [ ] Apply the solution planned in the `generateTenantToken` method.

brunoocasali:
For future readers: I opted to remove the tag/release because they were just an attempt to release, and I think this would be harmless for the users. 
Sorry for the inconvenience. If you have any trouble with this, please open an issue :)

bidoubiwa:
We opted to wait for users needing this special helper before choosing the best option. Meanwhile, if some users are in an environment that do support [jwt-kit](https://github.com/vapor/jwt-kit) you can import the library and create the tokens with the library. Here is a [pseudo-code](https://docs.meilisearch.com/learn/security/tenant_tokens.html#generating-tenant-tokens-without-a-meilisearch-sdk) and the guide to help you along. Please mention you are interested in this feature if you do (even if you use the third party library yourself)!

ppamorim:
@brunoocasali I would do what @bidoubiwa suggested. Abstract the implementation and leave the user to decide how to use jwt-kit. The less dependencies in the project the better.

brunoocasali:
Can you enlighten me more about how do think it could be done here?

PR #297, opened 2022-05-08
Remove all related to the generateTenantToken method

brunoocasali:
Removes everything related to generateTenantToken in order to unlock new releases to this package.

Related to #296


brunoocasali:
Since this is something really impactful for the users, I'll not wait for your review @bidoubiwa, but I still would like to know your ideas about how to solve this #296 🙏 .

PR #310, opened 2022-08-02
Add TaskInfo

brunoocasali:
Add a new type to help translate a brand new task to the real task resource. This will be used by every kind of async operation.

### DIFF
--- a/Sources/MeiliSearch/Model/TaskInfo.swift
+++ b/Sources/MeiliSearch/Model/TaskInfo.swift
@@ -1,0 +1,33 @@
+import Foundation
+
+/**
+ `TaskInfo` instances represent the current transaction status, use the `taskUid` value to
+  verify the status of your transaction.
+ */
+public struct TaskInfo: Codable, Equatable {
+
+  // MARK: Properties
+
+  /// Unique ID for the current `TaskInfo`.
+  public let uid: Int
+
+  /// Unique ID for the current `TaskInfo`.
+  public let indexUid: String?
+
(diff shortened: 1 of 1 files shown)


bidoubiwa (commented):
Does this mean that in Swift case a route like addDocuments returns an `uid` and not a `taskUid` ?

brunoocasali:
> Does this mean that in Swift case a route like addDocuments returns an `uid` and not a `taskUid` ?

yes, for now! I still have a lot of PR's to create here haha

Sources/MeiliSearch/Model/TaskInfo.swift
```diff
@@ -0,0 +1,33 @@
...
+
+  /// Date when the task has been enqueued.
+  public let enqueuedAt: String
+
+  static func == (lhs: TaskInfo, rhs: Task) -> Bool {
+    lhs.uid == rhs.uid
+  }
+
+  public enum CodingKeys: String, CodingKey {
+    case uid = "taskUid", indexUid, status, type, enqueuedAt
```
bidoubiwa, line 31: Same here for the multiple lines. I think it is more readable but it is up to you

brunoocasali: If we can adopt the raywenderlich style guide, this decision should be covered: https://github.com/raywenderlich/swift-style-guide#example-definition (in this case since the keys share the same context is ok to keep them in one line :))

bidoubiwa (approved):
LGTM 🔥

PR #311, opened 2022-08-02
Drop dumps

brunoocasali:
- Remove `Dump` type.
- Remove `Client#getDumpStatus` method.
- Make `Client#createDump` to return a Result<TaskInfo> instead of `Dump`.


Tests/MeiliSearchUnitTests/DumpsTests.swift
```diff
@@ -16,16 +16,12 @@ class DumpsTests: XCTestCase {
...
 
     let json = """
-      {
-        "uid": "20200929-114144097",
-        "status": "in_progress",
-        "startedAt": "2021-06-01T14:43:39.392327Z"
-      }
-      """
+      { "taskUid": 278, "indexUid": null, "status": "enqueued", "type": "dumpCreation", "enqueuedAt": "2022-07-21T21:43:12.419917471Z" }
+    """
```
bidoubiwa, line 20: Do you prefer this compact formating ?

brunoocasali: Yeah, I'm evaluating the best way to maintain this kind of test, but I'm 100% sure the type of format will not impact the ability to read the tests. Btw in multiple cases, these data didn't make any difference in the result :|

bidoubiwa: Yes, I think it is more readable when it is in multiple lines but you decide!

bidoubiwa (approved):
LGTM 🔥

PR #312, opened 2022-08-02
Document changes

brunoocasali:
- Remove `GetParameters` type
- Add `DocumentsQuery` type to handle the document endpoints' parameters.
- Add `DocumentsResults` type to handle the object response in the GET /documents response.
- Add a way to choose which `fields` the Meilisearch engine will return via `documents.get`
- Change responses in the create/update/delete actions for documents to return `TaskInfo`.
- Create the Queryable protocol to handle the creation of query strings


Tests/MeiliSearchUnitTests/DocumentsTests.swift
```diff
@@ -148,13 +148,13 @@ class DocumentsTests: XCTestCase {
 
   func testUpdateDocuments() {
     let jsonString = """
-      {"uid": 0, "indexUid": "movies_test", "status": "enqueued", "type": "documentAddition", "enqueuedAt": "xxx" }
+      {"taskUid":0,"indexUid":"books_test","status":"enqueued","type":"documentAdditionOrUpdate","enqueuedAt":"2022-07-21T21:47:50.565717794Z"}
```
bidoubiwa, line 151: Im not sure about removing the spaces between the fields. Do you think it is better?

Tests/MeiliSearchUnitTests/DocumentsTests.swift
```diff
@@ -6,10 +6,10 @@ import Foundation
...
 // swiftlint:disable line_length
 private struct Movie: Codable, Equatable {
-  let id: Int
-  let title: String
-  let overview: String
-  let releaseDate: Date
+  let id: Int?
+  let title: String?
+  let overview: String?
+  let releaseDate: Date?
```
bidoubiwa, line 12: overview and releaseDate seem to be optional in some of your tests, but `id` and `title` are used everywhere. Do they have to be optional ?

Tests/MeiliSearchUnitTests/DocumentsTests.swift
```diff
@@ -217,34 +217,111 @@ class DocumentsTests: XCTestCase {
...
+      let decoder = JSONDecoder()
+      decoder.dateDecodingStrategy = .formatted(Formatter.iso8601)
+      let data = jsonString.data(using: .utf8)!
+      let stubMovie: Movie = try decoder.decode(Movie.self, from: data)
+      let identifier: String = "25684"
+
+      // Start the test with the mocked server
+      let expectation = XCTestExpectation(description: "Get Movies document")
+
+      self.index.getDocument(identifier, fields: ["title", "id"]) { (result: Result<Movie, Swift.Error>) in
```
bidoubiwa, line 241: Since your documents already only contain `title` and `id` this tests is not actually testing if `fields` is useful. Maybe only try by adding `id` or `title` to see if one of the fields is removed

Tests/MeiliSearchUnitTests/DocumentsTests.swift
```diff
@@ -217,34 +217,111 @@ class DocumentsTests: XCTestCase {
...
+        case .failure:
+          XCTFail("Failed to get Movies document")
+        }
+        expectation.fulfill()
+      }
+    } catch {
+      XCTFail("Failed to parse document")
+    }
+
+//    self.wait(for: [expectation], timeout: TESTS_TIME_OUT)
```
bidoubiwa, line 254: I think this should not be commented. Sorry if there is a reason

Tests/MeiliSearchUnitTests/DocumentsTests.swift
```diff
@@ -217,34 +217,111 @@ class DocumentsTests: XCTestCase {
...
+    // Prepare the mock server
+    session.pushData(jsonString)
+
+    // Start the test with the mocked server
+    let expectation = XCTestExpectation(description: "Get documents with parameters")
+
+    self.index.getDocuments(params: DocumentsQuery(limit: 2, offset: 10)) { (result: Result<DocumentsResults<Movie>, Swift.Error>) in
+      switch result {
+      case .success:
+        XCTAssertEqual(self.session.nextDataTask.request?.url?.query, "limit=2&offset=10")
```
bidoubiwa, line 276: This test could be very useful on `getDocument` and the `fields` parameter

bidoubiwa (changes requested):

Sources/MeiliSearch/Documents.swift
```diff
@@ -16,18 +16,24 @@ struct Documents {
...
     _ uid: String,
     _ identifier: String,
+    fields: [String]? = nil,
     _ completion: @escaping (Result<T, Swift.Error>) -> Void)
   where T: Codable, T: Equatable {
-    let query: String = "/indexes/\(uid)/documents/\(identifier)"
+    var query: String = "/indexes/\(uid)/documents/\(identifier)"
+
+    if fields != nil {
+      query.append(fields?.joined(separator: ",") ?? "")
```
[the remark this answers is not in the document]
brunoocasali: You are correct, I was concerned about how Swift will parse JSON objects into structs that miss some fields. And my concern was confirmed because Swift breaks when the struct has missing fields in non-optional properties.

And then I totally miss to add this assertion: `XCTAssertEqual(self.session.nextDataTask.request?.url?.query, "fields=title,description")`

Sources/MeiliSearch/Documents.swift
```diff
@@ -36,30 +42,20 @@ struct Documents {
...
-        case .success(let data):
-          guard let data: Data = data else {
-            completion(.failure(MeiliSearch.Error.dataNotFound))
-            return
-          }
-          Documents.decodeJSON(data, completion: completion)
-
-        case .failure(let error):
-          completion(.failure(error))
+    request.get(api: "/indexes/\(uid)/documents\(params?.toQuery() ?? "")") { result in
```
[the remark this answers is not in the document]
brunoocasali: Good idea! :D

Sources/MeiliSearch/QueryParameters/DocumentsQuery.swift
```diff
@@ -0,0 +1,21 @@
+import Foundation
+
+public class DocumentsQuery: Queryable {
+  private var limit: Int?
+  private var offset: Int?
+  private var fields: [String]
+
+  init(limit: Int? = nil, offset: Int? = nil, fields: [String]? = nil) {
```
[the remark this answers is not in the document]
brunoocasali: Actually, there is a subtle difference between those implementations. Since the engine supports the `/documents?fields=` (without no value), which response with empty objects, I opted to receive as a default value `nil`, so if somebody wants that response they can do it :)

Sources/MeiliSearch/QueryParameters/DocumentsQuery.swift
```diff
@@ -0,0 +1,21 @@
...
+    self.offset = offset
+    self.limit = limit
+    self.fields = fields ?? []
+  }
+
+  internal func buildQuery() -> [String: Codable?] {
+    [
+      "limit": limit,
+      "offset": offset,
+      "fields": fields.isEmpty ? nil : fields.joined(separator: ",")
```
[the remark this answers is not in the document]
brunoocasali: I answered here: https://github.com/meilisearch/meilisearch-swift/pull/312#discussion_r936606588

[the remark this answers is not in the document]
brunoocasali: but indeed, there is an error in this logic, because the fields can be empty! It should be `"fields": fields != nil ? fields?.joined(separator: ",") : nil`

Sources/MeiliSearch/QueryParameters/Queryable.swift
```diff
@@ -0,0 +1,20 @@
+internal protocol Queryable {
+  func buildQuery() -> [String: Codable?]
+}
+
+extension Queryable {
+  func toQuery() -> String {
+    let query: [String: Codable?] = buildQuery()
+
+    let data = query.compactMapValues { $0 }
+      .sorted { $0.key < $1.key }
```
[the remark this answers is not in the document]
brunoocasali: For some weird reason, when I compact the values (remove nil's), Swift does not keep the order after that, then my assertions were flaky :/

Tests/MeiliSearchUnitTests/DocumentsTests.swift
brunoocasali: I just copy pasted a real response from the engine, I don't think there is much value about keeping or not those spaces.

Tests/MeiliSearchUnitTests/DocumentsTests.swift
brunoocasali: I didn't get it, because this is a unit test, so I control how the body will be retrieved, can you explain more this situation??

Tests/MeiliSearchUnitTests/DocumentsTests.swift
brunoocasali: I got this when it is not commented.
image

Tests/MeiliSearchUnitTests/DocumentsTests.swift
brunoocasali: yeah, agreed!

Tests/MeiliSearchUnitTests/DocumentsTests.swift
bidoubiwa: Interesting 😂 Maybe by taking inspiration from the other tests we can figure out? I have no idea why it does that. Maybe it is because of the use `do` ?

brunoocasali:
@bidoubiwa I made a PR with all the changes you requested #321 (I'll add all the requested changes in this PR)

bidoubiwa (approved):
LGTM 🔥

Tests/MeiliSearchUnitTests/DocumentsTests.swift
brunoocasali: It can be! But since this does not make a difference in multiple test suite executions, I could remove it without regrets!

Tests/MeiliSearchUnitTests/DocumentsTests.swift
brunoocasali: Yep I did that in other test :)

PR #313, opened 2022-08-02
Keys changes

brunoocasali:
- Add `Key#name` property & Create a struct `KeyUpdateParams` to update keys
- Replace `Task` with `TaskInfo` in the keys async occurrences
- Add `KeysQuery` type to handle the keys endpoints parameters

### DIFF
--- a/Sources/MeiliSearch/Model/KeyParams.swift
+++ b/Sources/MeiliSearch/Model/KeyParams.swift
@@ -5,17 +5,42 @@ import Foundation
  */
 public struct KeyParams: Codable, Equatable {
   public let description: String
+  public var name: String?
+  public var uid: String?
   public let actions: [String]
   public let indexes: [String]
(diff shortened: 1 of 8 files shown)


Sources/MeiliSearch/Model/KeyParams.swift
```diff
@@ -5,17 +5,42 @@ import Foundation
  */
 public struct KeyParams: Codable, Equatable {
   public let description: String
```
bidoubiwa, line 7: I think description is [optional as well](https://github.com/meilisearch/meilisearch-swift/pull/312#pullrequestreview-1060092053)

brunoocasali:
@bidoubiwa I made a PR with all the changes you requested #321 (I'll add all the requested changes in this PR)

bidoubiwa (approved):
LGTM 🔥

PR #314, opened 2022-08-02
Indexes changes

brunoocasali:
- Add `IndexesQuery` type to handle the indexes endpoints parameters.
- Add `IndexesResults` type to handle the object response in the `GET /indexes` response.
- Change responses in the create/update/delete actions for documents to return `TaskInfo`.

### DIFF
--- a/Sources/MeiliSearch/Indexes.swift
+++ b/Sources/MeiliSearch/Indexes.swift
@@ -101,15 +101,21 @@ public struct Indexes {
    completes the query request. It returns a `Result` object that contains `[Index]`
    value if the request was successful or `Error` if a failure occurred.
    */
-  public static func getAll(config: Config, _ completion: @escaping (Result<[Index], Swift.Error>) -> Void) {
-    Request(config).get(api: "/indexes") { result in
+  public static func getAll(config: Config, params: IndexesQuery? = nil, _ completion: @escaping (Result<Results<Index>, Swift.Error>) -> Void) {
+    Request(config).get(api: "/indexes", param: params?.toQuery()) { result in
       switch result {
       case .success(let result):
         guard let result: Data = result else {
           completion(.failure(MeiliSearch.Error.dataNotFound))
           return
         }
-        Indexes.decodeJSONArray(data: result, config: config, completion)
+
+        do {
+          let indexes: Results<Index> = try Constants.customJSONDecoder.decode(Results<Index>.self, from: result)
+          completion(.success(indexes))
+        } catch let error {
+          completion(.failure(error))
+        }
       case .failure(let error):
         completion(.failure(error))
       }
@@ -129,7 +135,7 @@ public struct Indexes {
     uid: String,
     primaryKey: String? = nil,
     config: Config,
-    _ completion: @escaping (Result<Task, Swift.Error>) -> Void) {
+    _ completion: @escaping (Result<TaskInfo, Swift.Error>) -> Void) {
     let payload = CreateIndexPayload(uid: uid, primaryKey: primaryKey)
     let data: Data
     do {
@@ -143,8 +149,8 @@ public struct Indexes {
       switch result {
       case .success(let data):
       do {
-        let result: Task = try Constants.customJSONDecoder.decode(
-          Task.self,
+        let result: TaskInfo = try Constants.customJSONDecoder.decode(
+          TaskInfo.self,
           from: data)
           completion(.success(result))
       } catch {
@@ -166,7 +172,7 @@ public struct Indexes {
    */
   public func update(
     primaryKey: String,
-    _ completion: @escaping (Result<Task, Swift.Error>) -> Void) {
+    _ completion: @escaping (Result<TaskInfo, Swift.Error>) -> Void) {
     let payload = UpdateIndexPayload(primaryKey: primaryKey)
     let data: Data
     do {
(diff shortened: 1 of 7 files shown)


Sources/MeiliSearch/Indexes.swift
```diff
@@ -101,15 +101,21 @@ public struct Indexes {
    completes the query request. It returns a `Result` object that contains `[Index]`
    value if the request was successful or `Error` if a failure occurred.
    */
-  public static func getAll(config: Config, _ completion: @escaping (Result<[Index], Swift.Error>) -> Void) {
-    Request(config).get(api: "/indexes") { result in
+  public static func getAll(config: Config, params: IndexesQuery? = nil, _ completion: @escaping (Result<Results<Index>, Swift.Error>) -> Void) {
```
bidoubiwa, line 104: Are we keeping `Result<Index>` and not `IndexResults` (like `DocumentsResult`  until the `GET /indexes` route has specific response fields that differ from other routes?

Sources/MeiliSearch/Indexes.swift
```diff
@@ -346,9 +352,10 @@ public struct Indexes {
    */
   public func getDocument<T>(
     _ identifier: String,
+    fields: [String]? = nil,
     _ completion: @escaping (Result<T, Swift.Error>) -> Void)
   where T: Codable, T: Equatable {
-    self.documents.get(self.uid, identifier, completion)
+    self.documents.get(self.uid, identifier, fields: fields, completion)
```
bidoubiwa, line 358: Should this not be added in the `Documents` PR ? (I can close my eyes if you want 😂)

Sources/MeiliSearch/Indexes.swift
```diff
@@ -376,10 +383,10 @@ public struct Indexes {
    failure occured.
    */
   public func getDocuments<T>(
-    options: GetParameters? = nil,
-    _ completion: @escaping (Result<[T], Swift.Error>) -> Void)
+    params: DocumentsQuery? = nil,
+    _ completion: @escaping (Result<DocumentsResults<T>, Swift.Error>) -> Void)
   where T: Codable, T: Equatable {
-    self.documents.getAll(self.uid, options, completion)
+    self.documents.getAll(self.uid, params: params, completion)
```
bidoubiwa, line 389: Same as above

Sources/MeiliSearch/Indexes.swift
```diff
@@ -392,7 +399,7 @@ public struct Indexes {
    */
   public func deleteDocument(
     _ documentId: String,
-    _ completion: @escaping (Result<Task, Swift.Error>) -> Void) {
+    _ completion: @escaping (Result<TaskInfo, Swift.Error>) -> Void) {
```
bidoubiwa, line 402: Same as above, should this not be added in the `Tasks` PR ?

Sources/MeiliSearch/Indexes.swift
```diff
@@ -560,7 +568,7 @@ public struct Indexes {
    value. If the request was sucessful or `Error` if a failure occured.
    */
   public func resetSynonyms(
-    _ completion: @escaping (Result<Task, Swift.Error>) -> Void) {
+    _ completion: @escaping (Result<TaskInfo, Swift.Error>) -> Void) {
```
bidoubiwa, line 571: All the `Task` to `TaskInfo` changes in the completion parameter are already changed in this pr #316

Sources/MeiliSearch/Indexes.swift
```diff
@@ -881,22 +889,6 @@ public struct Indexes {
...
-      let rawIndexes: [Index] = try Constants.customJSONDecoder.decode([Index].self, from: data)
-      var indexes: [Index] = []
-      for rawIndex in rawIndexes {
-        indexes.append(Index(uid: rawIndex.uid, primaryKey: rawIndex.primaryKey, createdAt: rawIndex.createdAt, updatedAt: rawIndex.updatedAt))
-      }
-      completion(.success(indexes))
-    } catch {
-      completion(.failure(error))
-    }
-  }
```
bidoubiwa, line 899: bye bye ❤️

Sources/MeiliSearch/Model/IndexesResults.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+
+/**
+ `IndexesResults` is a wrapper used in the indexes routes to handle the returned data.
+ */
+
+public struct IndexesResults<T: Decodable & Encodable & Equatable>: Codable, Equatable {
```
bidoubiwa, line 7: I see that there is a `IndexesResults` but somewhere above you used `Results<Index>`

Sources/MeiliSearch/Indexes.swift
brunoocasali: dammit, I forgot that yes close your eyes ahhaha!

Sources/MeiliSearch/Indexes.swift
brunoocasali: Not really. These small chunks are very time-consuming to create a PR with only these changes. Sorry about that!

Sources/MeiliSearch/Model/IndexesResults.swift
brunoocasali: Nice catch, I totally forgot them! :P

bidoubiwa (approved):
LGTM 🔥

PR #315, opened 2022-08-02
Tasks changes

brunoocasali:
- Create `TasksQuery` to handle query string for /tasks calls
- Add `TasksResults` type to handle the object response in the GET /tasks response.

### DIFF
--- a/Sources/MeiliSearch/Client.swift
+++ b/Sources/MeiliSearch/Client.swift
@@ -164,6 +164,14 @@ public struct MeiliSearch {
     self.tasks.waitForTask(task: task, options: options, completion)
   }
 
+  public func waitForTask(
+    task: TaskInfo,
+    options: WaitOptions? = nil,
+    _ completion: @escaping (Result<Task, Swift.Error>
+  ) -> Void) {
+    self.tasks.waitForTask(taskUid: task.uid, options: options, completion)
+  }
+
   // MARK: Tasks
 
  /**
@@ -181,15 +189,17 @@ public struct MeiliSearch {
   }
 
   /**
-   Get all tasks.
+   List tasks based on an optional pagination criteria
 
    - parameter completion: The completion closure used to notify when the server
+   - parameter params: A TasksQuery object with pagination metadata.
    completes the query request, it returns a `Result` object that contains `Key` value.
    If the request was sucessful or `Error` if a failure occured.
    */
   public func getTasks(
+    params: TasksQuery? = nil,
     _ completion: @escaping (Result<Results<Task>, Swift.Error>) -> Void) {
-    self.tasks.getAll(completion)
+    self.tasks.getAll(params: params, completion)
   }
 
   // MARK: Keys
--- a/Sources/MeiliSearch/QueryParameters/TasksQuery.swift
+++ b/Sources/MeiliSearch/QueryParameters/TasksQuery.swift
@@ -1,0 +1,30 @@
+import Foundation
+
+public class TasksQuery: Queryable {
+  private var from: Int?
+  private var limit: Int?
+  private var next: Int?
+  private var types: [String]
+  var indexUid: [String]
+  private var status: [String]
+
+  init(limit: Int? = nil, from: Int? = nil, next: Int? = nil, status: [String]? = nil, types: [String]? = nil, indexUid: [String]? = nil) {
+    self.from = from
+    self.limit = limit
+    self.next = next
+    self.status = status ?? []
+    self.types = types ?? []
+    self.indexUid = indexUid ?? []
+  }
+
+  internal func buildQuery() -> [String: Codable?] {
+    [
+      "limit": limit,
+      "from": from,
+      "next": next,
+      "type": types.isEmpty ? nil : types.joined(separator: ","),
+      "status": status.isEmpty ? nil : status.joined(separator: ","),
+      "indexUid": indexUid.isEmpty ? nil : indexUid.joined(separator: ",")
+    ]
+  }
+}
--- a/Sources/MeiliSearch/Tasks.swift
+++ b/Sources/MeiliSearch/Tasks.swift
@@ -21,14 +21,6 @@ struct Tasks {
       get(path: "/tasks/\(taskUid)", completion)
   }
 
-  // Get on index
-  func get(
-    indexUid: String,
-    taskUid: Int,
-    _ completion: @escaping (Result<Task, Swift.Error>) -> Void) {
-      get(path: "/indexes/\(indexUid)/tasks/\(taskUid)", completion)
-  }
-
   private func get (
     path: String,
     _ completion: @escaping (Result<Task, Swift.Error>) -> Void) {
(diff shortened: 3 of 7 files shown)


Sources/MeiliSearch/Client.swift
```diff
@@ -164,6 +164,14 @@ public struct MeiliSearch {
     self.tasks.waitForTask(task: task, options: options, completion)
   }
 
+  public func waitForTask(
```
bidoubiwa, line 167: I think you forgot to a comment above explaining the function and its parameters

Sources/MeiliSearch/QueryParameters/TasksQuery.swift
```diff
@@ -0,0 +1,30 @@
...
+  private var status: [String]
+
+  init(limit: Int? = nil, from: Int? = nil, next: Int? = nil, status: [String]? = nil, types: [String]? = nil, indexUid: [String]? = nil) {
+    self.from = from
+    self.limit = limit
+    self.next = next
+    self.status = status ?? []
+    self.types = types ?? []
+    self.indexUid = indexUid ?? []
+  }
```
bidoubiwa, line 18:
```suggestion
  init(limit: Int? = nil, from: Int? = nil, next: Int? = nil, status: [String]? = [], types: [String]? = [], indexUid: [String]? = []) {
    self.from = from
    self.limit = limit
    self.next = next
    self.status = status
    self.types = types
    self.indexUid = indexUid
  }
```

would this work?

Sources/MeiliSearch/QueryParameters/TasksQuery.swift
```diff
@@ -0,0 +1,30 @@
...
+
+  internal func buildQuery() -> [String: Codable?] {
+    [
+      "limit": limit,
+      "from": from,
+      "next": next,
+      "type": types.isEmpty ? nil : types.joined(separator: ","),
+      "status": status.isEmpty ? nil : status.joined(separator: ","),
+      "indexUid": indexUid.isEmpty ? nil : indexUid.joined(separator: ",")
```
bidoubiwa, line 27: Is this ternary useful since you are safe guarding above?

Sources/MeiliSearch/Tasks.swift
```diff
@@ -49,21 +41,34 @@ struct Tasks {
...
+    params: TasksQuery? = nil,
     _ completion: @escaping (Result<Results<Task>, Swift.Error>) -> Void) {
-      getAll(path: "/tasks", completion)
+      getAll(path: "/tasks", params: params, completion)
   }
 
   // get all on index
   func getAll(
     uid: String,
+    params: TasksQuery? = nil,
```
bidoubiwa, line 52: since the prototype is the same as the next overloaded function, how are we sure that we enter this function only when `getAll` is called from `indexes` ?

Sources/MeiliSearch/Tasks.swift
```diff
@@ -49,21 +41,34 @@ struct Tasks {
...
     _ completion: @escaping (Result<Results<Task>, Swift.Error>) -> Void) {
-      getAll(path: "/indexes/\(uid)/tasks", completion)
+      var query: TasksQuery?
+
+      if params != nil {
+        params?.indexUid.append(uid)
```
bidoubiwa, line 57: In case we enter this function through the `client.getTasks` and not via the `index.getTasks` the indexUid would be overwritten

Tests/MeiliSearchIntegrationTests/TaskTests.swift
```diff
@@ -57,31 +57,36 @@ class TasksTests: XCTestCase {
...
-      case .success:
-        addDocExpectation.fulfill()
-      case .failure(let error):
-        dump(error)
-        XCTFail("Failed to create index")
-        addDocExpectation.fulfill()
-      }
-    }
-    self.wait(for: [addDocExpectation], timeout: TESTS_TIME_OUT)
+    let expectation = XCTestExpectation(description: "Add documents")
```
bidoubiwa, line 60: not sure the description is still relevant here

Tests/MeiliSearchUnitTests/TasksTests.swift
```diff
@@ -0,0 +1,52 @@
...
+      }
+      """
+
+    // Prepare the mock server
+    session.pushData(jsonString)
+
+    // Start the test with the mocked server
+    let expectation = XCTestExpectation(description: "Get keys with parameters")
+
+    self.client.getTasks(params: TasksQuery(limit: 20, from: 5, next: 98, types: ["indexCreation"])) { result in
```
bidoubiwa, line 33: Maybe it is worth trying out with `indexUid: ["something"]` as well to ensure that the method called from the client is entering the correct`getAll`

Sources/MeiliSearch/Client.swift
brunoocasali: All the descriptions of the new classes/types/methods will be done after these PRs being merged :)

Sources/MeiliSearch/QueryParameters/TasksQuery.swift
brunoocasali: It works if I do some changes in the `buildQuery()` method, but I think its better with your suggestion!

Sources/MeiliSearch/QueryParameters/TasksQuery.swift
brunoocasali: After your suggestion I had to change to this: `"indexUid": indexUid != nil ? indexUid?.joined(separator: ",") : nil`

Sources/MeiliSearch/QueryParameters/TasksQuery.swift
brunoocasali: update: it does not work, because Meilisearch engine does not support `status=` or `type=` in the query string.

a validation like this will not work `status != nil ? status?.joined(separator: ",") : nil,` because since status is optional I will have to unwrap it before using it.

Sources/MeiliSearch/Tasks.swift
brunoocasali: When we call: `self.tasks.getAll(params: params, completion)` in the line 202 (from Client.swift) we are redirected to the `public` `getAll` function in the line 43. Then inside this function, we call the `private` `getAll` function, which just receives the params and makes the request.


When we call `self.tasks.getAll(uid: self.uid, params: params, completion)` in the line 476 (from Indexes.swift) we are redirected to the `public` `getAll` function in the line 51 not the public function in the line 43.

brunoocasali: To prevent those kind of misunderstandings I will rename this function :)

Sources/MeiliSearch/Tasks.swift
brunoocasali: I've answered here: https://github.com/meilisearch/meilisearch-swift/pull/315#discussion_r936755487 from my POV this cannot happen! :)

Tests/MeiliSearchIntegrationTests/TaskTests.swift
brunoocasali: CTRL+C CTRL+V spotted!

Tests/MeiliSearchUnitTests/TasksTests.swift
brunoocasali: The same here: https://github.com/meilisearch/meilisearch-swift/pull/315#discussion_r936756875

Sources/MeiliSearch/Tasks.swift
bidoubiwa: my bad I did not see they had different params

Sources/MeiliSearch/Client.swift
bidoubiwa: They will not be in #321 ?

bidoubiwa (approved):
bors merge

PR #316, opened 2022-08-02
Fix HTTP methods & Replace Task to TaskInfo

brunoocasali:
- Fix HTTP methods 
- Replace `Task` to `TaskInfo`

### DIFF
--- a/Sources/MeiliSearch/Settings.swift
+++ b/Sources/MeiliSearch/Settings.swift
@@ -41,7 +41,7 @@ struct Settings {
   func update(
     _ uid: String,
     _ setting: Setting,
-    _ completion: @escaping (Result<Task, Swift.Error>) -> Void) {
+    _ completion: @escaping (Result<TaskInfo, Swift.Error>) -> Void) {
 
     let data: Data
     do {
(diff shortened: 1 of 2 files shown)


Sources/MeiliSearch/Settings.swift
```diff
@@ -347,21 +347,21 @@ struct Settings {
   func updateDistinctAttribute(
     _ uid: String,
     _ distinctAttribute: String,
-    _ completion: @escaping (Result<Task, Swift.Error>) -> Void) {
+    _ completion: @escaping (Result<TaskInfo, Swift.Error>) -> Void) {
 
     let data: Data
     do {
-      data = try Constants.customJSONEecoder.encode(DistinctAttributePayload(distinctAttribute: distinctAttribute))
+      data = try JSONEncoder().encode(distinctAttribute)
```
bidoubiwa, line 354: Why were we able to change `Constants.customJSONEecoder.` here but not in the other methods?

brunoocasali: I think it's because this distinct-attribute setting just wants a simple string and the other methods require a complex object/array

bidoubiwa (approved):
LGTM 🔥

PR #319, opened 2022-08-02
Remove unused code in KeyTests & Add a cleanup step before KeyTests

brunoocasali:
- Remove all keys before starting a test to prevent order-dependency in the tests
- Make `description` an optional value

### DIFF
--- a/Sources/MeiliSearch/Model/KeyParams.swift
+++ b/Sources/MeiliSearch/Model/KeyParams.swift
@@ -4,7 +4,7 @@ import Foundation
  `KeyParams` contains all the parameters to create an API key.
  */
 public struct KeyParams: Codable, Equatable {
-  public let description: String
+  public var description: String?
   public var name: String?
   public var uid: String?
   public let actions: [String]
--- a/Tests/MeiliSearchIntegrationTests/KeysTests.swift
+++ b/Tests/MeiliSearchIntegrationTests/KeysTests.swift
@@ -5,69 +5,93 @@ import Foundation
 // swiftlint:disable force_try
 class KeysTests: XCTestCase {
   private var client: MeiliSearch!
-  private var key: String = ""
   private var session: URLSessionProtocol!
 
   // MARK: Setup
 
   override func setUp() {
     super.setUp()
+
     session = URLSession(configuration: .ephemeral)
     client = try! MeiliSearch(host: "http://localhost:7700", apiKey: "masterKey", session: session)
-    let keyExpectation = XCTestExpectation(description: "Get all keys")
 
-    self.client.getKeys { result in
+    // remove all keys to keep a clean state
+    self.client.getKeys(params: KeysQuery(limit: 100, offset: 0)) { result in
       switch result {
       case .success(let keys):
-        if keys.results.count > 0 {
-          let key = keys.results.first
-          if let firstKey: Key = key {
-            self.key = firstKey.key
+        keys.results.forEach {
+          self.client.deleteKey(key: $0.uid) { result in
+            switch result {
+            case .success:
+              ()
+            case .failure(let error):
+              dump(error)
+              XCTFail("Failed to delete key")
+            }
           }
-        } else {
-          XCTFail("Failed to get keys")
         }
-        keyExpectation.fulfill()
       case .failure(let error):
         dump(error)
-        XCTFail("Failed to get keys")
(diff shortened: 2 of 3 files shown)


Sources/MeiliSearch/Model/KeyParams.swift
```diff
@@ -4,7 +4,7 @@ import Foundation
  `KeyParams` contains all the parameters to create an API key.
  */
 public struct KeyParams: Codable, Equatable {
-  public let description: String
+  public var description: String?
```
bidoubiwa, line 7: Just out of curiosity, why are some var and some let? We cant change `actions` ? Is this because of `updateKey` ?

Tests/MeiliSearchIntegrationTests/KeysTests.swift
```diff
@@ -5,69 +5,93 @@ import Foundation
...
       case .success(let keys):
-        if keys.results.count > 0 {
-          let key = keys.results.first
-          if let firstKey: Key = key {
-            self.key = firstKey.key
+        keys.results.forEach {
+          self.client.deleteKey(key: $0.uid) { result in
+            switch result {
+            case .success:
+              ()
```
bidoubiwa, line 26: is this the same as `keyExpectation.fulfill()` ?

Tests/MeiliSearchIntegrationTests/KeysTests.swift
```diff
@@ -5,69 +5,93 @@ import Foundation
...
-          XCTFail("Failed to get keys")
         }
-        keyExpectation.fulfill()
       case .failure(let error):
         dump(error)
-        XCTFail("Failed to get keys")
-        keyExpectation.fulfill()
+        XCTFail("Failed to retrieve and delete all keys")
       }
     }
```
bidoubiwa, line 37: I think 
```
self.wait(for: [keyExpectation], timeout: TESTS_TIME_OUT)
```

might be missing

Tests/MeiliSearchIntegrationTests/KeysTests.swift
```diff
@@ -5,69 +5,93 @@ import Foundation
...
   func testGetKey() {
     let keyExpectation = XCTestExpectation(description: "Get one key")
 
-    self.client.getKey(key: self.key) { result in
+    self.client.createKey(KeyParams(actions: ["*"], indexes: ["*"], expiresAt: nil)) { result in
       switch result {
-      case .success(let key):
-        XCTAssertNotNil(key.description)
+      case .success(let createdKey):
+        self.client.getKey(key: createdKey.uid) { result in
```
bidoubiwa, line 73: maybe `key` should be renamed `keyOrUid` ? As it seems we are not suggesting `uid` as a valid option with this parameter name

Sources/MeiliSearch/Model/KeyParams.swift
brunoocasali: Correct!

Tests/MeiliSearchIntegrationTests/KeysTests.swift
brunoocasali: When we define a `switch`, we should provide an exhaustive list of possibilities, such as success and failure. The problem is that Swift requires me to provide at least a statement when defining a `case.` Since this method is meant to be a setup and I'm not interested in the success, I could provide the `()` empty function, but I would like to know if there is another way to handle that.

Tests/MeiliSearchIntegrationTests/KeysTests.swift
brunoocasali: I'm not sure because I fulfilling the expectation in every case.

Tests/MeiliSearchIntegrationTests/KeysTests.swift
bidoubiwa: Even if you are not interested in the success is `.fulfill()` not used to stop the wait of the expectation?

bidoubiwa (approved):
LGTM 🔥

Tests/MeiliSearchIntegrationTests/KeysTests.swift
brunoocasali: yeah, but that's the point, this code is written inside of the setup(), so it's not meant to be a "test" so in other words, no expectation should be needed 😬

PR #364, opened 2023-03-02
Fix tests with v0.30 Meilisearch

brunoocasali:
Just fix the tests, and the implementation will be done in other PRs


alallema (approved):
LGTM!
But why create a PR to fix the tests instead of creating a PR to update the task filters for:
- Rename status to statuses
- Rename type to types
- Rename indexUid to indexUids
?

brunoocasali:
> LGTM! But why create a PR to fix the tests instead of creating a PR to update the task filters for:
> 
> * Rename status to statuses
> * Rename type to types
> * Rename indexUid to indexUids
>   ?

It is easier to the reviewers!

PR #410, opened 2023-09-16
Add async overload for search API

### LINKED ISSUE #332: Add async/await support
**Description**
Since server-side Swift projects like Vapor now support async/await APIs, it would nice if meilisearch-swift also added this.

**Basic example**
As opposed to the completion parameter style currently used, an async/await call would look like so:

    let result = await index.updateDocuments(documents: [/** data **/], primaryKey: "id")
    switch result {
    case .success:
        // ....
    case .failure(let error):
        // ....
    }

**Proposed solution**
I have already created extensions in my local project to support some of the methods that I use. For example `updateDocuments()` is covered by this implementation:

    public func updateDocuments<T>(
      documents: [T],
      encoder: JSONEncoder? = nil,
      primaryKey: String? = nil) async -> Result<TaskInfo, Swift.Error> where T: Encodable {
          return await withCheckedContinuation { continuation in
              self.updateDocuments(documents: documents, encoder: encoder, primaryKey: primaryKey) { res in
                  continuation.resume(returning: res)
              }
          }
    }

Proposal is to add these bridging methods for _all_ methods on `Indexes` in perhaps an `Indexes+async.swift` extension. Additionally unit tests will need to be added to cover async calls.

Sherlouk:
# Pull Request

⚠️ I've marked this as draft as it current works to provide an example of how we could add async/await support to the library. I am more than happy to expand this to **all** currently supported APIs.

## Related issue

Relates to #332 (doesn't close as only adds support for one API)

## What does this PR do?
- Adds an async/await override to an existing search API providing users with more choice.

## PR checklist

### DIFF
--- a/Sources/MeiliSearch/Indexes.swift
+++ b/Sources/MeiliSearch/Indexes.swift
@@ -26,7 +26,7 @@ public struct Indexes {
   private let documents: Documents
 
   // Search methods
-  private let search: Search
+  fileprivate let search: Search
 
   // Settings methods
   private let settings: Settings
@@ -1098,6 +1098,26 @@ public struct Indexes {
     ) {
       self.uid = uid
       self.primaryKey = primaryKey
+    }
+  }
+}
+
+// MARK: Swift Concurrency (Async/Await)
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension Indexes {
+  /**
+   Search in the index.
+   
+   - Parameter searchParameters: Options on search.
+   - Throws: Error if a failure occurred.
+   - Returns: On completion if the request was successful a `Searchable<T>` instance is returned containing the values.
+   */
+  public func search<T: Codable & Equatable>(_ searchParameters: SearchParameters) async throws -> Searchable<T> {
+    try await withCheckedThrowingContinuation { continuation in
+      self.search.search(self.uid, searchParameters) { result in
+        continuation.resume(with: result)
+      }
     }
   }
 }
--- a/Sources/MeiliSearch/Indexes.swift
+++ b/Sources/MeiliSearch/Indexes.swift
@@ -26,7 +26,7 @@ public struct Indexes {
   private let documents: Documents
 
   // Search methods
-  fileprivate let search: Search
+  internal let search: Search
 
   // Settings methods
   private let settings: Settings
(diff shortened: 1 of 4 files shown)


Sources/MeiliSearch/Indexes.swift
```diff
@@ -26,7 +26,7 @@ public struct Indexes {
   private let documents: Documents
 
   // Search methods
-  private let search: Search
+  fileprivate let search: Search
```
aronbudinszky, line 29: This change is actually not required - `private` will work just as fine.

aronbudinszky (approved):
Aside from the minor comment regarding `fileprivate` all looks good to me!

I have not actually tested it though, but assume it does work given that the unit test passes.

Sources/MeiliSearch/Indexes.swift
Sherlouk: Let me double check this, I thought I had a compiler error when I tested (due to use of an extension) but will double check and revert if possible.

In respect of "I have not actually tested it though" - I'm using this in a production application and is working fine for me 🙂 Though I'll keep an eye on it!

curquiza:
Thank you @aronbudinszky for reviewing and @Sherlouk for the PR

Can you confirm there is no breaking for the user? 😊 
Is this change worth adding some documentation to the README? Or is ti really "usage transparent"?

Sherlouk:
This is not a breaking change, no.
I do think it's worth updating documentation with a quick example, but it is also obvious when using.

Sherlouk:
I've updated documentation, tweaked implementation, and opened up the PR. This does only add support for one API (search), but if we're happy then I can either:

1. Merge this PR, and then open a new one with support for all other APIs
2. Keep this PR open, until I've added support for all other APIs
3. Merge this PR, but open support for each other API one PR at a time (as I work on them with other features).

Clem, thoughts?

curquiza:
Let's go with #1, not with #3 indeed 😄

curquiza (approved):
bors merge

Sources/MeiliSearch/Indexes.swift
aronbudinszky: Yeah that's fair, agreed. (If there is value in hiding these properties internally we could still use protocols for that - though not sure if it is necessary.)

Only final point that I'd make then is that I would not explicitly state `internal` - given that this is the default anyway and not used on other internal properties.

aronbudinszky:
Btw retroactively confirming there is no breaking change. :)

Sources/MeiliSearch/Indexes.swift
Sherlouk: Fair point, that's a personal preference but deviates from standard practice for the project. I'll amend in a follow up PR

PR #413, opened 2023-09-17
Add support for search ranking score

### LINKED ISSUE #398: [v1.3] Display hits ranking scores
⚠️ This issue is generated, it means the examples and the namings do not necessarily correspond to the language of this repository.
Also, if you are a maintainer, please add any clarification and instructions about this issue.

Sorry if this is already wholly/partially implemented. Feel free to let me know about the state of this issue in the repo.

Related to https://github.com/meilisearch/integration-guides/issues/280

----

This issue is divided into two sections, first, you need to make the implementation, and second, you must update the code-samples (no one likes outdated docs, right?).

### New implementation

Related to:
- issue: https://github.com/meilisearch/meilisearch/issues/3771
- discussion: https://github.com/meilisearch/product/discussions/594 
- spec: https://github.com/meilisearch/specifications/pull/252

This feature aims to return ranking details for each document to understand and tweak the score of the documents more efficiently.

Ensure the SDKs can handle the new search parameter `showRankingScore`. Also, ensure the SDK can handle the `_rankingScore` attribute in the matched `hits`.

### Code samples

Inside of this file: `.code-samples.meilisearch.yml`:

1. Create a new entry with this key `search_parameter_guide_show_ranking_score_1` containing a call to the `search` method using the new behavior from the index `movies` using `showRankingScore=true` and a `q` with `'dragon'`.
Add a new code sample to the `.code-samples.meilisearch.yml` file:

Use this as a reference _if the previous description was not helpful_:

```
search_parameter_guide_show_ranking_score_1: |-
  POST 'http://localhost:7700/indexes/movies/search' 
  with data: { 
    "q": "dragon", "showRankingScore": true
  }
```

TODO:
- [ ] Add the ability receive a new param in the `search` method called `showRankingScore`.
- [ ] Add the ability handle the new `_rankingScore` key/value in the search hits' response.
- [ ] Update the code-samples accordingly
- [ ] Add integration tests

Sherlouk:
# Pull Request

## Related issue
Fixes #398

## What does this PR do?
- Add the ability receive a new param in the search method called showRankingScore.
- Add the ability handle the new _rankingScore key/value in the search hits' response.
- Add integration tests
- Update the code-samples accordingly

⚠️  **This is a breaking change**.

I'd love to hear some thoughts from others on how we may be able to avoid this kind of breaking change, but with #398 and #399 new values are being added to the "hit" object. The hit object is currently a generic provided by the user of the library, and as such it's impossible for us to add new attributes to that.

In this PR I've tried to get around this by creating our own encapsulation called "SearchHit" which itself holds a reference to the actual result but also any additional attributes that Meilisearch itself returns. In order to minimise the impact on users of the library I have implemented a "dynamic member lookup" with a keypath, this continues to support (in a completely type safe and compiler safe way) dot notation to variables within the hit result. Which means `hits[0].title` would still return the title on the hit, even though architecturally it's address is `hits[0].result.title`. While this does work for the vast majority of use cases, we are unable to support the case where a user tries to cast a hit to a type, you can see this in our tests where you can no longer cast `hits[0]` to a "Book", rather it is now a "SearchHit<Book>", but again... you can access variables the same and no other changes are needed to the code.

Importantly though this workaround with dynamic member lookups is limited to the capabilities of keypaths, which as documented officially by Apple [here](https://github.com/apple/swift/blob/main/test/attr/attr_dynamic_member_lookup.swift#L671-L672) does not include methods. As such where users have functions within their models, they would have to call that through `hits[0].result.someFunction()`.

I do personally think this is the best compromise. It is a breaking change, but most will not be impacted in any meaningful way. If others have a better idea though I'd love to hear it. This PR lays up the foundations for #399 which would be significantly easier once this is merged. But let's discuss.

## PR checklist

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -169,7 +169,7 @@ async_guide_canceled_by: |-
     }
   }
 search_parameter_guide_hitsperpage_1: |-
-  let searchParameters = SearchParameters.query("", hitsPerPage: 15)
+  let searchParameters = SearchParameters(query: "", hitsPerPage: 15)
   client.index("movies").search(searchParameters) { (result: Result<Searchable<Movie>, Swift.Error>) in
     switch result {
     case .success(let searchResult):
@@ -179,7 +179,7 @@ search_parameter_guide_hitsperpage_1: |-
     }
   }
 search_parameter_guide_page_1: |-
-  let searchParameters = SearchParameters.query("", page: 15)
+  let searchParameters = SearchParameters(query: "", page: 15)
   client.index("movies").search(searchParameters) { (result: Result<Searchable<Movie>, Swift.Error>) in
     switch result {
     case .success(let searchResult):
@@ -194,6 +194,16 @@ search_parameter_guide_attributes_to_search_on_1: |-
     switch result {
     case .success(let searchResult):
         print(searchResult)
+    case .failure(let error):
+        print(error)
+    }
+  }
+search_parameter_guide_show_ranking_score_1: |-
+  let searchParameters = SearchParameters(query: "dragon", showRankingScore: true)
+  client.index("movies").search(searchParameters) { (result: Result<Searchable<Movie>, Swift.Error>) in
+    switch result {
+    case .success(let searchResult):
+        print(searchResult.rankingScore)
     case .failure(let error):
         print(error)
     }
--- a/Sources/MeiliSearch/Model/SearchParameters.swift
+++ b/Sources/MeiliSearch/Model/SearchParameters.swift
@@ -62,6 +62,9 @@ public struct SearchParameters: Codable, Equatable {
   /// Whether to return the raw matches or not.
   public let showMatchesPosition: Bool?
 
+  /// Whether to return the search ranking score or not.
+  public let showRankingScore: Bool?
+
   // MARK: Initializers
 
   public init(
@@ -81,7 +84,8 @@ public struct SearchParameters: Codable, Equatable {
     filter: String? = nil,
     sort: [String]? = nil,
     facets: [String]? = nil,
-    showMatchesPosition: Bool? = nil) {
+    showMatchesPosition: Bool? = nil,
+    showRankingScore: Bool? = nil) {
     self.query = query
     self.offset = offset
     self.limit = limit
@@ -99,6 +103,7 @@ public struct SearchParameters: Codable, Equatable {
     self.sort = sort
     self.facets = facets
     self.showMatchesPosition = showMatchesPosition
+    self.showRankingScore = showRankingScore
   }
 
   // MARK: Query Initializers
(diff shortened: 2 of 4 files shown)


curquiza:
@Sherlouk thank for your PR
Multiple points
- There are now git conflict, can you fix them? 😄 
- Can you detail what your breaking is to have more context: what the users had to do before, and what should they do now?
- You have to know this SDK has really a low usage, so we can allow breaking. Of course we do our best to avoid them, especially if it concerns the main features (search and add documents)

Sherlouk:
Rebase completed, tests updated, and all running locally.

## Breaking Change

This is a breaking change affecting the results returned by use of the `search(_:)` function.

Previously, you would get back a response of `Searchable<T>` where "T" is a generic type representing the model of the document stored in MeiliSearch. `Searchable` includes an array called "hits" - the search results: an array of documents matching the query parameters.

With this PR, instead of "hits" being an array of documents, it is now an array of `SearchHit<T>`'s. You must now access the document contents by using `hits[0].document` instead of simply `hits[0]`. This encapsulating type allows us to add metadata unique to the hit but that is not part of the document itself (including the ranking score, and ranking stats, and any future expansion).

For convenience, this PR includes a dynamic member lookup. This isn't strictly necessary, but it does reduce (and in many cases completely avoids) any errors caused by this PR. It has some limitations, which I'll explain with some code snippets:

```swift
let result: Searchable<Book> = ... // unchanged

// accessing members (variables) on the document continues to work without change, thanks to the dynamic member lookup
print(result.hits[0].title) // VALID

// ---

// the result hit is no longer an instance of "T"
let book: Book = result.hits[0] // NEW ERROR
print(book.title)

let book: SearchHit<Book> = result.hits[0] // FIXED
print(book.title)

let book = result.hits[0] // ALSO FIXES ERROR
print(book.title)

let book: Book = result.hits[0].document // ALSO FIXES ERROR
print(book.title)

// ---

// functions do not work with dynamic member lookup
result.hits[0].getStock() // NEW ERROR
result.hits[0].document.getStock() // WITH FIX
```

Worst case, all errors caused by this PR simply need the developer to add `.document` to their usage as seen with the last example above. Best case, it continues to work with no changes necessary thanks to the dynamic member lookup.

curquiza:
Thank you @Sherlouk 
Following this breaking, do we need to update some examples the README or in the code samples file like [this one](https://github.com/meilisearch/meilisearch-swift/blob/4c6d2f4deaacfe4b67e213db71994466e647a8a3/.code-samples.meilisearch.yaml#L1046) (`getting_started_add_documents_md`)?

Sherlouk:
It's worth noting (though not necessarily required for this PR) that I do plan on doing a full review of code samples and documentation to improve adoption of the project. There are others that are outdated (but none because of this PR that I can see).

curquiza:
>It's worth noting (though not necessarily required for this PR) that I do plan on doing a full review of code samples and documentation to improve adoption of the project. There are others that are outdated (but none because of this PR that I can see).

Good to know! I will open an issue regarding it: I have scripts to know which code samples are missing and which ones should be removed 😊

curquiza (approved):
bors merge

PR #426, opened 2023-09-28
(Quick Wins) Fix Xcode 15, Reduce Search Requirements, Update README

### LINKED ISSUE #218: Add a search example with `filter` in README
⚠️ This issue is generated, please adapt the example to the repository language.

Related to: https://github.com/meilisearch/integration-guides/issues/136

In the README, we want to add the `Custom Search With Filters` sub section in the `Getting Started` section to make the users are aware of filtering.

This should
- be placed right after the `Custom Search` sub section
- contain the following content:

---------- beginning of the content

If you want to enable filtering, you must add your attributes to the `filterableAttributes` index setting.

--- *Add an example corresponding to the current repository. See this example in JS:*
```js
await index.updateAttributesForFaceting([
    'id',
    'genres'
  ])
```
--- *end of example*

You only need to perform this operation once.

Note that MeiliSearch will rebuild your index whenever you update `filterableAttributes`. Depending on the size of your dataset, this might take time. You can track the process using the [update status](https://docs.meilisearch.com/reference/api/updates.html#get-an-update-status).

Then, you can perform the search:

--- *Add an example corresponding to the current repository. See this example in JS:*
```js
await index.search(
  'wonder',
  {
    filter: ['id > 1 AND genres = Action']
  }
)
```
--- *end of example*

```json
{
  "hits": [
    {
      "id": 2,
      "title": "Wonder Woman",
      "genres": ["Action","Adventure"]
    }
  ],
  "offset": 0,
  "limit": 20,
  "nbHits": 1,
  "processingTimeMs": 0,
  "query": "wonder"
}
```

------------- end of the content

📣 Example of PR: https://github.com/meilisearch/meilisearch-js/pull/1059/files

### LINKED ISSUE #420: Search 'T' must conform to Encodable and Equatable
While perhaps not always onerous to do, not everybody needs nor wants their documents to conform to Encodable and/or Equatable to perform a simple search using the package.

It may be prudent to look into whether this requirement is necessary, and whether we can get away with simply just `T: Decodable` instead to make the package as easy to use as possible.

### LINKED ISSUE #421: Unsupported URL when Schema Unspecified
Strictly speaking, the package calls for the "host" which under spec doesn't include the schema. If I specify a URL without the schema, only when requesting a resource do you get an error:

> `meiliSearchCommunicationError(message: "unsupported URL", url: "<google.com>/indexes/stations/search")`

URLs are validated when the client is initialised, so it's unclear why this gets so far through the stack before failing.

We should be more resilient to different kinds of URL inputs.
We should also provide more documentation (in README and inline).
We should do more thorough validation on client initialisation.

Sherlouk:
# Pull Request

## Related issue
Fixes #421
Fixes #420 
Fixes #218 

## What does this PR do?
Nothing huge, new, or breaking. Just three real small issues on the project that this PR can quickly close down including a bug introduced in Xcode 15 - the test now passes locally but Xcode on GitHub defaults to Xcode 14.3 which doesn't have the issue.

## PR checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,18 +95,6 @@ curl -L https://install.meilisearch.com | sh
 
 NB: you can also download Meilisearch from **Homebrew** or **APT** or even run it using **Docker**.
 
-### Import
-
-After installing `meilisearch-swift`, you must import it into your application.
-
-```swift
-import MeiliSearch
-
-let client = try! MeiliSearch(host: "http://localhost:7700")
-```
-
-When initialising the MeiliSearch client, you must provide a fully qualified URL including scheme.
-
 ## 🎬 Getting started
 
 To do a simple search using the client, you can create a Swift script like this:
@@ -117,6 +105,7 @@ To do a simple search using the client, you can create a Swift script like this:
     import MeiliSearch
 
     // Create a new client instance of Meilisearch.
+    // Note: You must provide a fully qualified URL including scheme.
     let client = try! MeiliSearch(host: "http://localhost:7700")
 
     struct Movie: Codable, Equatable {
--- a/README.md
+++ b/README.md
@@ -183,6 +183,48 @@ Since Meilisearch is typo-tolerant, the movie `philadelphia` is a valid search r
 
 > Note: All package APIs support closure-based results for backwards compatibility. Newer async/await variants are being added under [issue 332](https://github.com/meilisearch/meilisearch-swift/issues/332).
 
+#### Custom Search With Filters <!-- omit in toc -->
+
+If you want to enable filtering, you must add your attributes to the `filterableAttributes` index setting.
+
+```swift
+index.updateFilterableAttributes(["id", "genres"]) { result in
+    // Handle Result in Closure
+}
+```
+
+You only need to perform this operation once.
+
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -95,6 +95,18 @@ curl -L https://install.meilisearch.com | sh
 
 NB: you can also download Meilisearch from **Homebrew** or **APT** or even run it using **Docker**.
 
+### Import
```
curquiza, line 98: What is the point of this section? The import is already explicit in the Getting Started section, right?
Maybe I miss a swift usage.

My point is to avoid having a too "heavy" README, and make it simple and user-friendly, so avoid the duplication of information.

Sherlouk: Happy to remove or rewrite, the main line from this is trying to call out this:

> When initialising the MeiliSearch client, you must provide a fully qualified URL including scheme.

This has the potential to catch people out, although we do now have some additional protection.

Sherlouk: I've removed the section in favour of a single line within the Getting Started code sample. Hopefully this is more appropriate for you 🙂

curquiza (approved):
thank you!

bors merge

PR #427, opened 2023-10-01
Generate Async Overloads for All Public Functions

### LINKED ISSUE #332: Add async/await support

Sherlouk:
# Pull Request

## Related issue
Fixes #332 

This is NOT a breaking change.

## What does this PR do?
- Adds async/await overloads to all public interfaces.

The contents of these files has been automatically generated using [Sourcery](https://github.com/krzysztofzablocki/Sourcery) using [this template](https://gist.github.com/Sherlouk/fa3e49e24f9be7232c642b945177b05e) I made. It does not include documentation, I feel like this is an okay compromise at this point in time and is something that can be incremented on in the future, potentially when we move over entirely. Async/Await functions sit in Xcode along with the normal closure based methods and so is easy to see the documentation inline anyway.

The template has been shared to hopefully make it easier to update when new APIs are added in the future, keeping maintenance down.

With the previous PR (#410) I called the underlying implementation directly, this leads to unnecessary duplication with other functions. Instead I call the non-async function guaranteeing consistent functionality, and no unnecessary duplication. I also feel like this avoids the need for excessive and additional test coverage, so long as the build compiles then the implementation is working as expected and tests give us no extra confidence that it's working as expected. We'll keep the one for search to demonstrate.

@aronbudinszky Are you happy with this approach too?

## PR checklist

### DIFF
--- a/Sources/MeiliSearch/Async/Client+async.swift
+++ b/Sources/MeiliSearch/Async/Client+async.swift
@@ -181,8 +181,8 @@ extension MeiliSearch {
   /**
    See `isHealthy(_:)`
    */
-  public func isHealthy() async throws -> Bool {
-    try await withCheckedThrowingContinuation { continuation in
+  public func isHealthy() async -> Bool {
+    await withCheckedContinuation { continuation in
       self.isHealthy { result in
         continuation.resume(returning: result)
       }
--- a/Sources/MeiliSearch/Async/Client+async.swift
+++ b/Sources/MeiliSearch/Async/Client+async.swift
@@ -102,6 +102,28 @@ extension MeiliSearch {
   }
 
   /**
+   See `cancelTasks(filter:completion:)`
+   */
+  public func cancelTasks(filter: CancelTasksQuery) async throws -> TaskInfo {
+    try await withCheckedThrowingContinuation { continuation in
+      self.cancelTasks(filter: filter) { result in
+        continuation.resume(with: result)
+      }
+    }
+  }
+
+  /**
+   See `deleteTasks(filter:completion:)`
+   */
+  public func deleteTasks(filter: DeleteTasksQuery) async throws -> TaskInfo {
+    try await withCheckedThrowingContinuation { continuation in
+      self.deleteTasks(filter: filter) { result in
+        continuation.resume(with: result)
+      }
+    }
+  }
+
+  /**
    See `getKeys(params:_:)`
    */
   public func getKeys(params: KeysQuery? = nil) async throws -> KeysResults {
--- a/Sources/MeiliSearch/Async/Client+async.swift
+++ b/Sources/MeiliSearch/Async/Client+async.swift
@@ -1,0 +1,213 @@
+import Foundation
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension MeiliSearch {
+  /**
+   See `createIndex(uid:primaryKey:_:)`
+   */
+  public func createIndex(uid: String, primaryKey: String? = nil) async throws -> TaskInfo {
+    try await withCheckedThrowingContinuation { continuation in
+      self.createIndex(uid: uid, primaryKey: primaryKey) { result in
+        continuation.resume(with: result)
+      }
+    }
+  }
+
+  /**
+   See `getIndex(_:_:)`
+   */
+  public func getIndex(_ uid: String) async throws -> Index {
+    try await withCheckedThrowingContinuation { continuation in
+      self.getIndex(uid) { result in
+        continuation.resume(with: result)
+      }
+    }
+  }
+
+  /**
+   See `getIndexes(params:_:)`
+   */
+  public func getIndexes(params: IndexesQuery? = nil) async throws -> IndexesResults {
+    try await withCheckedThrowingContinuation { continuation in
+      self.getIndexes(params: params) { result in
+        continuation.resume(with: result)
+      }
+    }
+  }
+
+  /**
+   See `updateIndex(uid:primaryKey:_:)`
+   */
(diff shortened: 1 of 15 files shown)


Sources/MeiliSearch/Async/Client+async.swift
```diff
@@ -0,0 +1,213 @@
...
+  /**
+   See `isHealthy(_:)`
+   */
+  public func isHealthy() async -> Bool {
+    await withCheckedContinuation { continuation in
+      self.isHealthy { result in
+        continuation.resume(returning: result)
+      }
+    }
+  }
```
Sherlouk, line 190: This implementation is very slightly different to all others - this is because the underlying implementation does not return a result, but rather returns `false` if an error is identified.

As such we use a checked continuation (no throwing), and forward the bool response verbatim.

Works the same, just no throwing.

curquiza (approved):
I'm ok with this PR, thank you!

Let's wait for the @aronbudinszky feedback before merging.
If we don't get any answer, we will merge it.

Sources/MeiliSearch/Async/Client+async.swift
aronbudinszky: Nice.

However this does point to an important omission that I would say blocks this PR currently - the lack of tests.

I think one way to test things might be to check that the async results are equivalent to the non-async ones - this would avoid duplicating tests but still ensure our assumptions about success, failure, and the compatibility with the old-style api.

(If you have time cool, if not I can probs take a look at it as well in the next few days.)

aronbudinszky (changes requested):

aronbudinszky:
Sorry somehow missed your comment about tests above - though I _kinda_ agree that parity is _theoretically_ assured, I still think that tests are important because going to async can sometimes cause weird issues and/or there can be _some_ exceptions to how things are implemented exactly.

Also in case there are any test coverage requirements (which probs there should be) this would help with those stats... :)

aronbudinszky:
@curquiza Just to confirm, this is *not* a breaking change, so if we choose to add tests in an upcoming PR I'm also happy to merge this in.

aronbudinszky:
(On a side note as I see there are [plans](https://github.com/meilisearch/meilisearch-swift/issues/405) to not allow untested code to be merged.)

curquiza:
Thanks @aronbudinszky for your reactivity.
About tests you are right, and I would rather we have the tests in the same PR than the addition, if possible 😊

Sherlouk:
Spent a little bit of time looking into the tests, and so wanted to see what folks thought a good balance would be for now:

1. Duplicate all tests so every functionality has an async/await test and a closure based test
2. Replace all closure based tests with async/await tests (under the basis that every async implementation calls the closure based one so in terms of test coverage it's identical and all integrations are touched)
3. Create minified tests to async/await to simply assert a request to the correct API URL path but no assertions on response type (which is covered by existing closure tests).

I'm wary of adding too much duplication and maintenance overhead with this PR, especially until I get to rewriting some of the unit and integration tests to reduce prose. 

My opinion would be continuing with number two for now, moving us in the strategic direction of defaulting to async/await, reducing complexity in tests (with use of expectations), and by extension improving performance of tests. It still guarantees a 100% test coverage for public APIs.

Sherlouk:
I've pushed up a couple examples of my proposal to aide in discussion, the new tests are running about 80% faster locally. 

I also believe them to be far more readable, though this will be especially prevalent when I get to the more complex tests with many nested levels of callbacks.

Tests/MeiliSearchUnitTests/Utils.swift
```diff
@@ -6,3 +6,8 @@ import XCTest
 @testable import MeiliSearch
 
 public let TESTS_TIME_OUT = 10.0
+
+func decodeJSON<T: Decodable>(from string: String) throws -> T {
+  let data = Data(string.utf8)
+  return try Constants.customJSONDecoder.decode(T.self, from: data)
+}
```
Sherlouk, line 13: This isn't a necessary change for the sake of async/await switching, but utilising a common extension here allows us to remove more force unwraps increasing safety throughout tests, and shorter call site increases readability elsewhere.

curquiza:
Sorry for the delay @Sherlouk 

I'm ok to go with solution 2

Let me know when you think this is ready to be merged for you!

Sherlouk:
Thanks Clem, will give @aronbudinszky some time to reply too to ensure we're all aligned and then will get it finished.

curquiza (approved):

Sherlouk:
bors merge

Sherlouk:
Going to send this through for now to unlock further work as this is a predecessor for quite a few bits. 

If Aron or any other consumer of the application has any opinions on how we can improve this epic then I'd love to hear it - feel free to open a new issue and we can iterate on this together!

PR #431, opened 2023-11-03
Allow Custom Headers

### LINKED ISSUE #290: Give the possibility to provide custom headers
As for #283 the different requests method provide `headers` as an optionnal parameter. 
Nonetheless, if we go up in the stack of calls, no headers are passed to these methods. We should be able to pass them.

For example here: 

https://github.com/meilisearch/meilisearch-swift/blob/431d8eefacd9d18e623d19f96cc1b110d675ddfa/Sources/MeiliSearch/Documents.swift#L16-L23

the `request.get` does not provide the headers that can be passed as an argument here

https://github.com/meilisearch/meilisearch-swift/blob/431d8eefacd9d18e623d19f96cc1b110d675ddfa/Sources/MeiliSearch/Request.swift#L31-L35

We should be able to provide the information. 

Two possible solutions `this.config` can contain the specific headers. In which case we can remove the `headers` params from the request methods. 

Or we provide them as an optional parameter for each method: 

```swift
 public func createIndex(
    uid: String,
    primaryKey: String? = nil,
    headers: [String: String] = [:],
    _ completion: @escaping (Result<Task, Swift.Error>) -> Void) {
    Indexes.create(uid: uid, primaryKey: primaryKey, config: self.config, completion)
  }
```
I prefer the first option but this also means an additional parameter during the initialisation of a Meilisearch client. 

```swift
  public init(
     host: String,
     apiKey: String? = nil,
     session: URLSessionProtocol? = nil,
     request: Request? = nil,
     headers: [String: String] = [:]) throws {
```

Sherlouk:
# Pull Request

## Related issue
Fixes #290

## What does this PR do?
- Adds a new parameter to initialiser to allow passing through of custom headers
- Custom header will not override API key which takes highest precedence
- Remove unused request parameter
- Remove unused header parameters in network functions

Though the public signature does change (for removal of 'request'), it was never possible to initialise your own Request due to no public initialiser being provided. As such it's impossible for any client to have this parameter in use, and thus this is **not** a breaking change.

## PR checklist

### DIFF
--- a/Sources/MeiliSearch/Client.swift
+++ b/Sources/MeiliSearch/Client.swift
@@ -33,14 +33,11 @@ public struct MeiliSearch {
    - parameter host:   The host to the Meilisearch http server.
    - parameter apiKey:    The authorization key to communicate with Meilisearch.
    - parameter session:   A custom produced `URLSessionProtocol`.
+   - parameter headers: A dictionary of custom headers passed with every API request.
    */
-  public init(host: String, apiKey: String? = nil, session: URLSessionProtocol? = nil, request: Request? = nil) throws {
-    self.config = try Config(host: host, apiKey: apiKey, session: session).validate()
-    if let request: Request = request {
-      self.request = request
-    } else {
-      self.request = Request(self.config)
-    }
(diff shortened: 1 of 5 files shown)


brunoocasali (approved):
LGTM!

Sherlouk:
bors merge

Issue #434, opened 2023-12-16
NIO Client Pitch


kaishin:
Hi @Sherlouk. I previously opened issue #91 to pitch adding support for NIO and server-side Swift projects. It was subsequently closed since I couldn't dedicate time to address the questions raised nor to open source the NIO package I mentioned. 

Since then, I managed to open source the NIO client implementation as [meilisearch-nio](https://github.com/kaishin/meilisearch-nio) and I have been using it for over 2 years now.

In the [original pitch](https://github.com/meilisearch/meilisearch-swift/issues/91#issuecomment-727932369) I brought attention to the amount of duplication between the two projects, especially the models and request payloads. I still think there is value in   either merging or making the NIO client depend on this repository, but that may require some potentially breaking changes.

Ideally, this repo can host the base models, the `URLSession` client for Darwin, and the NIO client as three separate packages. That would provide a lot of flexibility while keeping complexity at a minimum. 

I would love to hear your thoughts.

Sherlouk:
Hey Reda.

Pedro asked a really valuable question in the initial pitch:

> Is it possible to use [this custom URLSession](https://github.com/meilisearch/meilisearch-swift/blob/master/Sources/MeiliSearch/Config.swift#L61) from the Config initializer

In fact, this is the exact approach I personally took in my [meilisearch-vapor](https://github.com/Sherlouk/meilisearch-vapor/tree/main) project. Just like your proposal, it utilises NIO and AHC via Vapor. It could be made more generic to support any NIO implementation (though it wasn't my use case and so didn't build it). It has full Linux support and I've been using in production for a couple months now.

Why does this approach not work for what you're trying to achieve?

kaishin:
Hi Sherlouk,

Thank you for taking the time to reply. 

> Pedro asked a really valuable question in the initial pitch:

I initially didn't manage to look into Pedro's suggestion but looking at your vapor bridge I see the idea now. This approach does indeed solve the need to have separate clients for client-side apps and servers, especially if it's abstracted away like it is the case in your Vapor wrapper.

If you think this should be the recommended way to interface with MeiliSearch from server-side Swift, a mention of the Vapor repo in the README would be quite valuable instead of the current demo link. 

Out of curiosity, is this kind of request bridging common in the server-side Swift ecosystem ? I don't recall coming across it before.

---

[...]

PR #437, opened 2024-01-27
Extract models into a separate Core module

kaishin:
This PR is the first step of addressing #434 .


Package.swift
```diff
@@ -2,23 +2,64 @@
...
+      .target(name: MeiliSearchCore)
+    ]
+  ),
+  .target(
+    name: MeiliSearchCore, 
+    dependencies: []
+  ),
+  .testTarget(
+    name: "MeiliSearchUnitTests",
+    dependencies: ["MeiliSearch"]
```
Sherlouk, line 29: nit: where we reference target names, let's be consistent and use the static variable

```suggestion
    dependencies: [.target(MeiliSearch)]
```

Package.swift
```diff
@@ -2,23 +2,64 @@
...
+  )
+]
+
+#if os(Linux) || USE_NIO
+products.append(.library(name: MeiliSearchNIO, targets: [MeiliSearchNIO]))
+dependencies.append(.package(url: "https://github.com/swift-server/async-http-client.git", from: "1.19.0"))
+targets.append(
+  .target(
+    name: MeiliSearchNIO,
```
Sherlouk, line 42: major: this PR will currently fail to build as-is due to this target path not existing. shall we delete this entire Linux section until you contribute the NIO integration in your next PR?

Package.swift
```diff
@@ -2,23 +2,64 @@
...
+    name: "MeiliSearchUnitTests",
+    dependencies: ["MeiliSearch"]
+  ),
+  .testTarget(
+    name: "MeiliSearchIntegrationTests",
+    dependencies: ["MeiliSearch"]
+  )
+]
+
+#if os(Linux) || USE_NIO
```
Sherlouk, line 37: comment: I like the use of `USE_NIO` here as it'll allow macOS clients (for working on Vapor projects locally) to build the NIO integration. When we add the NIO integration (presumably the next PR), let's document how to use this.

Sources/MeiliSearchCore/PackageVersion.swift
```diff
@@ -1,18 +1,15 @@
 import Foundation
-#if canImport(FoundationNetworking)
-  import FoundationNetworking
-#endif
 
-internal struct PackageVersion {
+public struct PackageVersion {
   /// This is the current version of the meilisearch-swift package
-  private static let current = "0.15.0"
+  public static let current = "0.15.0"
```
Sherlouk, line 5: question: why has the access level of this variable changed? where else is it used?

Sherlouk, line 5: minor: let's make sure we update [the documentation](https://github.com/meilisearch/meilisearch-swift/blob/main/CONTRIBUTING.md#how-to-publish-the-release-) where this line of code is referenced (the path to this file)

Sherlouk, line 5: minor: let's make sure we update [the Podspec](https://github.com/meilisearch/meilisearch-swift/blob/main/MeiliSearch.podspec) where this line of code is referenced (the path to this file)

Tests/MeiliSearchIntegrationTests/DocumentsTests.swift
```diff
@@ -1,4 +1,5 @@
 @testable import MeiliSearch
+@testable import MeiliSearchCore
```
Sherlouk, line 2: question: do we require @testable for the core target? 99% of models within are public (applicable to all tests)

Package.swift
```diff
@@ -2,23 +2,64 @@
...
+
+var targets: [PackageDescription.Target] = [
+  .target(
+    name: MeiliSearch,
+    dependencies: [
+      .target(name: MeiliSearchCore)
+    ]
+  ),
+  .target(
+    name: MeiliSearchCore,
```
Sherlouk, line 24: comment: this new target should not break the Podspec as it does a global search for all Swift files within the Sources directory, however when we do add the NIO target it will fail to compile due to the lack of the dependency.

AHC (or NIO) is not available officially on CocoaPods, so this will be a SPM-only feature. I'm okay with that but we must document it and ensure that CocoaPods does continue to work (even if only using Foundation).

Sherlouk (changes requested):

kaishin:
@Sherlouk Thanks for the review! I wasn't expecting it yet as this is still a draft but happy to address the comments I hadn't planned to address.

Sherlouk:
> happy to address the comments I hadn't planned to address

No worries, thought I may as well share some guiding thoughts. Looks like great progress towards the shared goal 👍

Sherlouk:
Hey @kaishin - just wondering how you're feeling about this?

kaishin:
@Sherlouk The project where I used MeiliSearch got deprioritized during spring so I couldn't get myself to dedicate time to finish this, but that's gonna be changing this month so perhaps let's timebox this to August? If I don't get it done this month I am unlikely to finish it afterwards.

PR #455, opened 2024-12-09
Add KeyActions enum

### LINKED ISSUE #418: Make Key Actions Enum
Similar to #363, but for key actions.

* All values are available in documentation here: https://www.meilisearch.com/docs/reference/api/keys#actions
* Enum should use camelCase (so `documents.add` = `documentsAdd`)
* User should be able to provide unsupported action to array
* Response should be able to handle unsupported actions returned (minus compiler safety)

MiaKoring:
# Pull Request

## Related issue
Fixes #418 

## What does this PR do?
- add KeyAction enum with all supported cases and .unknown(String)
- extends KeyAction with an initializer with String
- extends KeyAction with a String representation of the enum value
- adds initializer with [KeyAction] instead of String for actions to KeyParams
- extends Key&KeyParams with value “enumActions” containing the actions as KeyAction representation
- 
## PR checklist
- [X] Have you read the contributing guidelines?
- [X] Have you made sure that the title is accurate and descriptive of the changes?


Sherlouk:
This is awesome, thanks! Will give it a full review shortly and look to get both merged in (in preparation for a release once a few more breaking changes are compiled).

Sources/MeiliSearch/Model/KeyParams.swift
```diff
@@ -67,3 +83,9 @@ public struct KeyUpdateParams: Codable, Equatable {
     }
   }
 }
+
+public extension KeyParams {
+  var enumActions: [KeyAction] {
+    self.actions.map { KeyAction($0) }
+  }
+}
```
Sherlouk, line 91: What are the thoughts on making this a computed variable as opposed to simply changing `actions` from a [String] to [KeyAction].

It's a breaking change, but I feel best to store as "typed" as possible.

Tests/MeiliSearchIntegrationTests/KeysTests.swift
```diff
@@ -171,6 +171,54 @@ class KeysTests: XCTestCase {
...
+        XCTAssertEqual(key.actions, keyParams.actions)
+        XCTAssertEqual(key.enumActions, keyParams.enumActions)
+        XCTAssertEqual(key.indexes, keyParams.indexes)
+        XCTAssertNotNil(key.expiresAt)
+        keyExpectation.fulfill()
+      case .failure(let error):
+        print(error.localizedDescription)
+        dump(error)
+        XCTFail("Failed to create a key")
+        keyExpectation.fulfill()
```
Sherlouk, line 196: I'm assuming this test is a "happy path", and as such we should remove this expectation fulfill, and the comment above which states "is expected to fail".

Sources/MeiliSearch/Model/KeyParams.swift
MiaKoring: I decided to do it that way, so it’d be an addition but not required. If you prefer it as replacement, I’m happy to change it

Sherlouk: I think given we're still getting the API in a solid place, I'd want a core principle of this library to be type safety -- and thus would encourage storing it as type safe solution as opposed to the string array.

This is a discussion not a decision though, so if you disagree let's chat about that.

Sherlouk: It's less of a concern and more an awareness that we're storing it in its most compressed state (a String) which delays the point at which it becomes typed.

It also requires users to know that the extension exists, where default behaviour will be to use the string variable.

I think by replacing the array type we ensure users are using the optimal approach by default, and have to go out of their way to use strings.

MiaKoring: I agree on that. I'm going to implement it later or in the next few days :)

Sherlouk (approved):
Will need to merge the other PR first @curquiza, but both should be good to go.

Let's not do a release yet though -- I'll work on some other tweaks soon as my cal frees up!

curquiza:
@MiaKoring 
can you fix the git conflict please? we had to merge https://github.com/meilisearch/meilisearch-swift/pull/459 first, sorry

Sherlouk:
Sorry @MiaKoring! With this PR we'll do both the MTask and KeyActions.

CI will be working once you bring in latest changes, and so can get it merged.

Thanks

Sherlouk (approved):
All looking good. Ignore the linter error, I'll tweak SwiftLint in a separate PR as it's a little sensitive right now.

curquiza:
@Sherlouk I cannot merge with the linter error. Do you want to open a PR to remove the linting issue? we will merge this PR after

curquiza (approved):
bors merge

